PR #3, opened 2025-05-24
Add Mkdocs plugins to pip install

darabos:
Sorry — since this runs on merge, I'm afraid it will be a few PRs before I get everything right.


darabos:
Wow it worked! https://lynxkite.github.io/lynxkite-2000/

PR #13, opened 2025-05-29
Fix test reliability issues

darabos:
Also fixes some real issues.

### DIFF
--- a/lynxkite-app/web/tests/lynxkite.ts
+++ b/lynxkite-app/web/tests/lynxkite.ts
@@ -126,26 +126,19 @@ export class Workspace {
     await this.page.mouse.up();
   }
 
-  async tryToConnectBoxes(sourceId: string, targetId: string) {
+  async connectBoxes(sourceId: string, targetId: string) {
     const sourceHandle = this.getBoxHandle(sourceId, "right");
     const targetHandle = this.getBoxHandle(targetId, "left");
+    await expect(sourceHandle).toBeVisible();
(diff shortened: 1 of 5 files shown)


lynxkite-app/web/tests/lynxkite.ts
```diff
@@ -132,10 +129,16 @@ export class Workspace {
   async connectBoxes(sourceId: string, targetId: string) {
     const sourceHandle = this.getBoxHandle(sourceId, "right");
     const targetHandle = this.getBoxHandle(targetId, "left");
+    await expect(sourceHandle).toBeVisible();
+    await expect(targetHandle).toBeVisible();
```
darabos, line 133: This line is the key. I'm not sure why we even see the target box without a handle, but if we don't wait for it to appear, sometimes the edge doesn't get created between them. (I can't reproduce this manually.)

darabos:
Wow this was an adventure! :sweat_smile:

In the end it looks like the issue is when you click too early on the output plug. Unfortunately, I can't find any outward sign of when is it too early. So we can't wait for it to be ready. I also can't figure out what goes wrong when it's too early. It's not simply the CRDT update cutting the connection -- I've done manual experiments with that and couldn't reproduce the issue.

What seems to work now (it passed 3 times!) is retrying the connection if it failed after 1 second. It doesn't make the test noticeably slower. Not my proudest work, but hopefully we will have reliable tests now.

PR #26, opened 2025-06-20
A box for combining Bundles

darabos:
@JMLizano take a look! I added the merge stuff as a new box to better show it off, but I think it should be combined with "Create graph". Except then I would rename it. Maybe just call it "Organize"? We already have a box called that, but that can be deleted. 😅 Sorry about the mess.


JMLizano:
This looks quite good thanks!

I had in mind something more generic though, allowing multi input for any kind of boxes and environments. But seems like we will need different implementations per environment. This is already useful as it is. I added some minimal testing, but I can't push to this PR, neither create a new PR, I am lacking permissions. I will wait for you to be back to give me those permissions!

PR #30, opened 2025-07-03
Add export functionality

BoraAy:
Add export functionality, so far I only added the same file formats as in importing.

### DIFF
--- a/lynxkite-graph-analytics/src/lynxkite_graph_analytics/lynxkite_ops.py
+++ b/lynxkite-graph-analytics/src/lynxkite_graph_analytics/lynxkite_ops.py
@@ -75,7 +75,7 @@ def import_file(
     return core.Bundle(dfs={table_name: df})
 
 
-@op("Export to File")
+@op("Export to file")
 def export_to_file(
     bundle: core.Bundle,
     *,


lynxkite-graph-analytics/src/lynxkite_graph_analytics/lynxkite_ops.py
```diff
@@ -75,6 +75,45 @@ def import_file(
     return core.Bundle(dfs={table_name: df})
 
 
+@op(
+    "Export to File",
+)
```
darabos, line 80:
```suggestion
@op("Export to file")
```

lynxkite-graph-analytics/src/lynxkite_graph_analytics/lynxkite_ops.py
```diff
@@ -75,6 +75,45 @@ def import_file(
...
+    else:
+        columns = [col.strip() for col in columns.split(",")]
+    if file_format == FileFormat.csv:
+        df.to_csv(filename, index=index, columns=columns)
+    elif file_format == FileFormat.json:
+        df.to_json(filename, orient="records", lines=True)
+    elif file_format == FileFormat.parquet:
+        df.to_parquet(filename, index=index)
+    elif file_format == FileFormat.excel:
+        df.to_excel(filename, index=index)
```
darabos, line 112: These formats ignore `columns`. That's a bit error prone -- you export a CSV without the secret stuff, it's all good, then you later change to JSON and suddenly the secrets are in the JSON.

- One option is to apply `columns` everywhere. This is not too hard. You can just do `df = df[columns]`.
- But it's even easier to just drop the `columns` parameter. The user can already remove columns easily.

lynxkite-graph-analytics/src/lynxkite_graph_analytics/lynxkite_ops.py
```diff
@@ -75,6 +75,45 @@ def import_file(
...
+    Args:
+        bundle: The bundle containing the DataFrame to export.
+        filename: The name of the file to export to.
+        file_format: The format of the file to export to. Defaults to CSV.
+        columns: Comma-separated list of columns to export.
+            Defaults to "<all>" which exports all columns.
+        index: Whether to include the DataFrame index in the exported file. Defaults to False.
+    """
+
+    df = bundle.dfs["result"]
```
darabos, line 100: What if I want to export a different table? Please make the table name a parameter. (A string is fine. A lot of boxes already take a table name parameter. At some point I will fix them all to offer a dropdown.)

darabos (commented):
Excellent, thanks! I think this list of file formats is enough. We can add others later on demand.

darabos:
Maybe you could add a test for this? If we have tests for import boxes (I'm not sure) then it could work similar to that.

How does this work when you re-run the workspace? Does it quietly overwrite the output file? (I think that's the best option. I don't want to have to manually delete it or change the name all the time.)

PR #32, opened 2025-07-03
Fix crash when collapsing table view box

darabos:
As @Mackofi discovered, closing/opening "View tables" boxes would lead to a crash in Workspace.tsx with the error:
```
Error: Not supported: reassigning object that already occurs in the tree.
```

This happens when instead of making a change inside a CRDT object, you replace the whole object. Based on that, this PR is quote logical, and it fixes the issue. But which line in particular fixes it? I'll add a comment.

### DIFF
--- a/lynxkite-app/web/src/workspace/Workspace.tsx
+++ b/lynxkite-app/web/src/workspace/Workspace.tsx
@@ -128,24 +128,32 @@ function LynxKiteFlow() {
           !Number.isNaN(ch.position.y)
         ) {
           getYjsDoc(state).transact(() => {
-            Object.assign(node.position, ch.position);
+            node.position.x = ch.position.x;
+            node.position.y = ch.position.y;
           });
         } else if (ch.type === "select") {
         } else if (ch.type === "dimensions") {
-          getYjsDoc(state).transact(() => Object.assign(node, ch.dimensions));
+          getYjsDoc(state).transact(() => {
+            node.width = ch.dimensions.width;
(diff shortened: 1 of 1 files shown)


lynxkite-app/web/src/workspace/Workspace.tsx
```diff
@@ -128,24 +128,32 @@ function LynxKiteFlow() {
...
-          getYjsDoc(state).transact(() => Object.assign(node.data, u));
+          getYjsDoc(state).transact(() => {
+            if (node.data.collapsed !== ch.item.data.collapsed) {
+              node.data.collapsed = ch.item.data.collapsed;
+            }
+            if (node.data.__execution_delay !== ch.item.data.__execution_delay) {
+              node.data.__execution_delay = ch.item.data.__execution_delay;
+            }
+            for (const [key, value] of Object.entries(ch.item.data.params)) {
+              if (node.data.params[key] !== value) {
```
darabos, line 152: It's this line! When closing/opening the box, the `collapsed` attribute is changed. But without this line we replace all the parameters.

I think I misunderstood the issue! Maybe replacing a CRDT object is okay (after all, that happens when you change a parameter). But you can't replace it with an object that's already in the CRDT document tree. So `params.foo = { bar }` is okay, but `params.foo = params.foo` is _not_ okay.

Still a bit of a mystery why we haven't seen this elsewhere.

Mackofi (approved):
These fixes do fix the crash on my end too. Thanks!

PR #39, opened 2025-07-09
Special characters in workspace name cause bugs

### LINKED ISSUE #34: Special characters in workspace name cause bugs
When giving workspaces names with special characters in them (e.g. `%` or `+`), they cause the workspace to break. For example:
- When you create a workspace with a name that contains `%` sign, then the site crashes. 
- When you give it a name like `test + test`, everything appears normal, but when you try to manually re-run the workspace, the workspace `test   test` will be the one running.

Mackofi:
Fixes #34.
The culprit was that we did not encode the URIs and I think I have fixed that in all places. I self tested these with these (` §'"+!%=(),.-~ˇ^˘°˛``˙´ `) special characters and all actions (create folder, create file, create workspace, and all of these nested) and they seemed to work fine, and I have not seen any crashes since.

### DIFF
--- a/lynxkite-app/web/src/Directory.tsx
+++ b/lynxkite-app/web/src/Directory.tsx
@@ -29,7 +29,7 @@ function EntryCreator(props: {
   onCreate: (name: string) => void;
 }) {
   const [isCreating, setIsCreating] = useState(false);
-  const [error, setError] = useState("");
+  const [nameValidationError, setError] = useState("");
 
   function validateName(name: string): boolean {
     if (name.includes("/")) {
@@ -58,7 +58,7 @@ function EntryCreator(props: {
           }}
         >
           <input
-            className={`input input-ghost w-full ${error ? "input-error" : ""}`}
+            className={`input input-ghost w-full ${nameValidationError ? "input-error" : ""}`}
             autoFocus
             type="text"
             name="entryName"
@@ -66,7 +66,7 @@ function EntryCreator(props: {
             onChange={(e) => validateName(e.target.value)}
             placeholder={`${props.label} name`}
           />
-          {error && (
+          {nameValidationError && (
             <div
               className="error-message"
               role="alert"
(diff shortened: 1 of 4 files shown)


Mackofi:
Okay, `#` and `?` don't work yet. Looking into how to fix it.

lynxkite-app/web/src/Directory.tsx
```diff
@@ -29,23 +29,53 @@ function EntryCreator(props: {
...
+      setError("Name cannot contain '/' characters");
+      return false;
+    }
+    if (name.trim() === "") {
+      setError("Name cannot be empty");
+      return false;
+    }
+    setError("");
+    return true;
+  }
```
darabos, line 45: Sounds good to me! I think this is in line with what Linux accepts as filename. (Linux also disallows the `null` character (`chr(0)`), but I don't think there's a risk of anyone typing that.)

lynxkite-app/web/src/Directory.tsx
```diff
@@ -29,23 +29,53 @@ function EntryCreator(props: {
   onCreate: (name: string) => void;
 }) {
   const [isCreating, setIsCreating] = useState(false);
+  const [error, setError] = useState("");
```
darabos, line 32: There are many things that could have an error on this page. Since this is displayed for the naming of new workspaces/folders, I would give it a more specific name.

lynxkite-app/web/src/Directory.tsx
```diff
@@ -67,13 +97,17 @@ export default function Directory() {
   const navigate = useNavigate();
 
   function link(item: DirectoryEntry) {
+    const encodedName = item.name
```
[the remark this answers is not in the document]
darabos: Agreed.

darabos (commented):
Awesome, thanks a lot!

lynxkite-app/web/src/Directory.tsx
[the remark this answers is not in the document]
Mackofi: Done!

lynxkite-app/web/src/Directory.tsx
Mackofi: Named it nameValidationError. Sounds descriptive to me.

PR #40, opened 2025-07-09
Composable pytorch layers

JMLizano:
**Add support to create hierarchical models by specifying Pytorch modules as inputs of other modules.**
I have done it by identifying a new type of node, the submodule. A submodule is a node in the workspace whose output is connected to an input of another node, and that input is of type `torch.nn.Module` or `list[torch.nn.Module]`.
A submodule doesn't add a Layer to the final model, instead it is stored on a new `submodoles` dict, to be later retrieved by the operation that uses it.

**Add support for `list` inputs in Pytorch environment.**
Copied over from https://github.com/biggraph/lynxkite-2000-enterprise/pull/220 , I think it makes sense to include this change here.

### DIFF
--- a/lynxkite-graph-analytics/src/lynxkite_graph_analytics/pytorch/pytorch_core.py
+++ b/lynxkite-graph-analytics/src/lynxkite_graph_analytics/pytorch/pytorch_core.py
@@ -2,6 +2,7 @@
 
 import copy
 import graphlib
+from typing import get_origin
 
 import pydantic
 from lynxkite.core import ops, workspace
@@ -228,6 +229,7 @@ class ModelBuilder:
             for i in v:
                 self.inv_dependencies[i].append(k)
         self.layers: list[Layer] = []
+        self.submodules: dict[str, torch.nn.Module] = {}
         # Clean up disconnected nodes.
         to_delete = set()
         for node_id in self.nodes:
@@ -235,10 +237,10 @@ class ModelBuilder:
             if title not in self.catalog:  # Groups and comments, for example.
                 to_delete.add(node_id)
                 continue
-            op = self.catalog[title]
-            if len(self.in_edges[node_id]) != len(op.inputs):  # Unconnected inputs.
-                to_delete.add(node_id)
-                to_delete |= self.all_upstream(node_id)
+            # op = self.catalog[title]
+            # if len(self.in_edges[node_id]) != len(op.inputs):  # Unconnected inputs.
+            #     to_delete.add(node_id)
+            #     to_delete |= self.all_upstream(node_id)
         for node_id in to_delete:
             del self.dependencies[node_id]
             del self.in_edges[node_id]
@@ -268,6 +270,10 @@ class ModelBuilder:
         t = node.data.title
         op = self.catalog[t]
         p = op.convert_params(node.data.params)
+        if self.is_submodule(node_id):
+            print(f"node id  {node_id} submodule: {self.is_submodule(node_id)}")
+            self.submodules[node_id] = self.run_op(node_id, op, p)
+            return
         match t:
             case "Repeat":
                 if node_id.startswith("END "):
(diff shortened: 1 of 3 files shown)


lynxkite-graph-analytics/src/lynxkite_graph_analytics/pytorch/pytorch_core.py
```diff
@@ -262,6 +284,23 @@ def all_downstream(self, node: str) -> set[str]:
...
 
+    def _is_submodule_node(self, node_id: str) -> bool:
+        """Returns True if the node is a submodule
+
+        A submodule is a node in the workspace whose output is connected to an input of another
+        node, and that input is of type torch.nn.Module or list[torch.nn.Module].
+        In other words, a submodule is a node that is used as a component (module) inside a
+        higher-level module, rather than being a top-level layer in the model sequence.
+        """
+        for output, connections in self.out_edges[node_id].items():
```
JMLizano, line 295: I have doubts about this, how should we treat multiple output nodes? This kind of relies on the node having a single output. Can a node be both a submodule and return a non submodule output at the same time?

I decided to not address this in this PR, and choose the easy way out :sweat_smile:

lynxkite-graph-analytics/src/lynxkite_graph_analytics/pytorch/pytorch_core.py
```diff
@@ -300,21 +339,71 @@ def run_node(self, node_id: str) -> None:
...
+            operation_inputs.append(input_edges)
         outputs = [_to_id(node_id, n.name) for n in op.outputs]
         if op.func == ops.no_op:
             module = torch.nn.Identity()
         else:
-            module = op.func(*inputs, **params)
-        return Layer(module, node_id, inputs, outputs)
+            module = op.func(*operation_inputs, **params)
+        # Sub-modules are not real inputs, they are just a way to build hierarchical models.
+        non_module_inputs = [
```
JMLizano, line 397: There is one thing that I have skipped for now. This implementation ignores submodules inputs, expecting the parent module to handle the inputs.
Based on that we should allow to include submodules without connected inputs in the UI, but currently nodes without inputs are deleted, so you just still need to connect some input to the submodule. It is not urgent, but it would be a nice QOL improvement.

lynxkite-graph-analytics/src/lynxkite_graph_analytics/pytorch/pytorch_ops.py
```diff
@@ -115,7 +115,7 @@ def dropout(x, *, p=0.0):
 
 
 @op("Linear", weights=True)
-def linear(x, *, output_dim=1024):
+def linear(x: torch.Tensor, *, output_dim=1024):
```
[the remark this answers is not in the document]
JMLizano: Right, this is still confusing for me :sweat_smile:

JMLizano:
> * I thought the submodule part would be left unparsed, and only parsed when it's passed to the module that uses it. Instead here it's eagerly parsed (turned into nn.Module). Maybe it doesn't matter.

This is the easiest/most efficient way I found, we were already traversing the nodes, so we do the rollup on the go, instead of adding an extra step in the `__init__`. But I am open for suggestions.

> * It looks like your code only supports submodules with a single layer. This is a big limitation that I'd like to remove. E.g. for Neural ODE we want to pass in an MLP.

Unless you mean something else, this already allows to hierarchically build submodules of several layers. For example:

 - GraphConv-1 is used as input of HeteroConv
 - GraphConv-2 is used as input of HeteroConv
 - HeteroConv is then used as an input for HeteroEncoder (made up for the example)

I have added a new test to showcase this.

JMLizano:
Yes, but you will need an extra box to combine the Linear layers, like Sequential.

image

But if what you expect is for those 3 boxes to get combined automatically in a submodule, then no.

JMLizano:
Ah I see, I think we have a different definition for submodules.

This PR solves the issue of one box (module) depending on other boxes (submodules) at model build time. My goal here was to solve the issue I had with the HeteroConv box, that needs other GraphConv boxes at build time.

But what you are explaning, seems like something similar to the hat box in old LynxKite?

lynxkite-graph-analytics/src/lynxkite_graph_analytics/pytorch/pytorch_core.py
JMLizano: Shouldn't we also add the condition that those other outputs are not used as submodules? Or is there a reason to enforce a single output to be connected?

Issue #29, opened 2025-07-03
PyKEEN integration


darabos:
@Mackofi has started working on integrating https://github.com/pykeen/pykeen. This is a big task and we will break it into smaller issues, but currently I don't even know what those small issues would be, so I'm filing this as a starting point.

- **Step 1 is writing a design document for the integration.** Plan out the kind of boxes we will have. I expect we could have a series of boxes in the LynxKite Graph Analytics environment that just wrap different aspects of PyKEEN. (E.g., load dataset, load model, create embeddings.)
- **Consider the PyTorch model environment.** Is there a part of PyKEEN that provides or uses PyTorch modules? If so, we could add boxes in this environment. (I expect we don't do this, but I haven't looked closely.)
- Even if we don't integrate inside the PyTorch model environment, we want to be able to **use existing embeddings for nodes and edges** in PyKEEN. E.g. I've created node embeddings with an LLM or with a domain-specific model like Geneformer. I want to pass these into PyKEEN to form the basis of the graph embedding.
- We also want to **use the PyKEEN embedding in LynxKite models**. Make sure the output of PyKEEN is in a format that can easily be fed into a PyTorch model.
- **Step 2: implement the design.**
- **Step 3: create demo applications.**

Mackofi:
I have written a draft-like design document (found [here](https://docs.google.com/document/d/1LQFC0kzm5ac508XQJcTki3O9tH3X4YunWlb-uFf3Ds4/edit?usp=sharing)). And I would like your comments/ideas/suggestions.

I have also started implementing a few easier looking and (in my opinion) straight-forward features on the branch 'szabo-pykeen'.

darabos:
> I have written a draft-like design document (found [here](https://docs.google.com/document/d/1LQFC0kzm5ac508XQJcTki3O9tH3X4YunWlb-uFf3Ds4/edit?usp=sharing)). And I would like your comments/ideas/suggestions.

Excellent! I've added some comments.

> I have also started implementing a few easier looking and (in my opinion) straight-forward features on the branch 'szabo-pykeen'.

Perfect! Yeah, avoid anything complex for starters. Just build whatever is simplest.

PR #42, opened 2025-07-10
Start of PyKEEN integration

Mackofi:
I have implemented the basic features PyKEEN provides (according to the [design document](https://docs.google.com/document/d/1LQFC0kzm5ac508XQJcTki3O9tH3X4YunWlb-uFf3Ds4)), but I am sure there are things that can and will be improved down the line.

Sadly we discovered that loading embeddings into PyKEEN models from other sources (as described in #29) is not possible the way we wanted it, as PyKEEN only supports edge types. So that feature will have to be skipped, unless we find a reasonable replacement.

I have also provided a workspace demonstrating the usage of PyKEEN, and I am currently working on an example that involves another machine learning task using the embeddings extracted from the trained PyKEEN model. But that is very much not ready yet.

The workspace is still work in progress and I will continue to iterate on it, but right now everything is in a good enough state for a draft.

### DIFF
--- a/examples/pykeen-testing.py
+++ b/examples/pykeen-testing.py
@@ -1,4 +1,4 @@
-from lynxkite.core import ops
+from lynxkite_core import ops
 from lynxkite_graph_analytics import core
 
 from lynxkite_graph_analytics.pykeen_ops import PyKEENModelWrapper
@@ -13,6 +13,7 @@ import torch
 op = ops.op_registration("LynxKite Graph Analytics")
 
 
+# TODO: make this into an actual op, I have plans see branch `szabo-pykeen-load-embeddings`
 @op("Load embedding into PyKEEN model")
 def load_pykeen_embeddings(
     dataset: core.Bundle,
@@ -101,6 +102,7 @@ def load_pykeen_embeddings(
         entity_to_id=entity_to_id,
         relation_to_id=relation_to_id,
         edges_data=edges_data,
+        seed=42,
     )
 
     bundle_dataset.other[save_as] = model_wrapper
--- a/lynxkite-graph-analytics/src/lynxkite_graph_analytics/core.py
+++ b/lynxkite-graph-analytics/src/lynxkite_graph_analytics/core.py
@@ -42,28 +42,6 @@ OtherName = typing.Annotated[str, {"format": "dropdown", "metadata_query": "[].o
 rendered as a dropdown in the frontend, listing the keys on the "other" part of the Bundle.
 The key is passed to the operation as a string."""
 
-PyTorchModelName = typing.Annotated[
-    str,
-    {
-        "format": "dropdown",
-        "metadata_query": "[].other.*[] | [?type == 'pytorch-model'].key",
-    },
-]
-"""A type annotation to be used for parameters of an operation. ModelName is
-rendered as a dropdown in the frontend, listing the models in the Bundle.
-The model name is passed to the operation as a string."""
-
-PyKEENModelName = typing.Annotated[
-    str,
-    {
-        "format": "dropdown",
-        "metadata_query": "[].other.*[] | [?type == 'pykeen-model'].key",
-    },
-]
-"""A type annotation to be used for parameters of an operation. PyKEENModelName is
-rendered as a dropdown in the frontend, listing the PyKEEN models in the Bundle.
-The model name is passed to the operation as a string."""
-
 # Parameter names in angle brackets, like <table_name>, will be replaced with the parameter
 # values. (This is not part of JMESPath.)
 # ColumnNameByTableName will list the columns of the DataFrame with the name
--- a/lynxkite-graph-analytics/src/lynxkite_graph_analytics/lynxkite_ops.py
+++ b/lynxkite-graph-analytics/src/lynxkite_graph_analytics/lynxkite_ops.py
@@ -135,9 +135,9 @@ def import_graphml(*, filename: str):
 
 @op("Graph from OSM", slow=True)
 def import_osm(*, location: str):
-    from osmnx import graph
+    import osmnx as ox
 
-    return graph.graph_from_place(location, network_type="drive")
+    return ox.graph.graph_from_place(location, network_type="drive")
 
 
 @op("Discard loop edges")
--- a/lynxkite-graph-analytics/src/lynxkite_graph_analytics/pykeen_ops.py
+++ b/lynxkite-graph-analytics/src/lynxkite_graph_analytics/pykeen_ops.py
@@ -1,9 +1,10 @@
 """PyKEEN operations."""
 
-from lynxkite.core import ops
+from lynxkite_core import ops
 from . import core
 from .pytorch.pytorch_core import _torch_save, _torch_load
 
+import typing
 import pandas as pd
 import enum
 from pykeen import models
@@ -15,6 +16,18 @@ from pykeen.triples import TriplesFactory
 
 
 op = ops.op_registration(core.ENV)
+
+
+PyKEENModelName = typing.Annotated[
+    str,
+    {
+        "format": "dropdown",
+        "metadata_query": "[].other.*[] | [?type == 'pykeen-model'].key",
+    },
+]
+"""A type annotation to be used for parameters of an operation. PyKEENModelName is
+rendered as a dropdown in the frontend, listing the PyKEEN models in the Bundle.
+The model name is passed to the operation as a string."""
 
 
 def mapped_triples_to_df(triples, entity_to_id, relation_to_id):
@@ -164,11 +177,13 @@ class PyKEENModel(str, enum.Enum):
     TransR = "TransR"
     TuckER = "TuckER"
 
-    def to_class(self, triples_factory: TriplesFactory, embedding_dim: int) -> models.Model:
+    def to_class(
+        self, triples_factory: TriplesFactory, embedding_dim: int, seed: int = 42
+    ) -> models.Model:
         return getattr(models, self.value)(
             triples_factory=triples_factory,
             embedding_dim=embedding_dim,
-            random_seed=42,
+            random_seed=seed,
         )
 
 
@@ -183,6 +198,7 @@ class PyKEENModelWrapper:
         entity_to_id: dict,
         relation_to_id: dict,
         edges_data: pd.DataFrame,
+        seed: int,
         trained: bool = False,
     ):
         self.model = model
@@ -192,6 +208,7 @@ class PyKEENModelWrapper:
         self.entity_to_id = entity_to_id
         self.relation_to_id = relation_to_id
         self.edges_data = edges_data
+        self.seed = seed
 
     def metadata(self) -> dict:
         return {
@@ -212,14 +229,8 @@ class PyKEENModelWrapper:
         return str(self.model)
 
     def __getstate__(self):
-        state = {
-            "trained": getattr(self, "trained", False),
-            "model_type": getattr(self, "model_type", None),
-            "embedding_dim": getattr(self, "embedding_dim", None),
-            "entity_to_id": getattr(self, "entity_to_id", None),
-            "relation_to_id": getattr(self, "relation_to_id", None),
-            "edges_data": getattr(self, "edges_data", None),
-        }
+        state = dict(self.__dict__)
+        del state["model"]
         if self.trained:
             state["model_state_dict"] = _torch_save(self.model.state_dict())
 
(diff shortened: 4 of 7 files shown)


lynxkite-graph-analytics/src/lynxkite_graph_analytics/pykeen_ops.py
```diff
@@ -0,0 +1,331 @@
...
+    if "model" in bundle.other:
+        model = bundle.other["model"]
+    else:
+        sampler = None
+        if model is PyKEENModel.RGCN and training_approach == TrainingType.sLCWA:
+            # Currently RGCN is the only model that requires a sampler and only when using sLCWA
+            sampler = "schlichtkrull"
+        model = model.to_class()
+    training_set = TriplesFactory.from_labeled_triples(
+        bundle.dfs["triples_train"][["head", "relation", "tail"]].values,
```
darabos, line 184: Take the train/test table names as parameters.

lynxkite-graph-analytics/src/lynxkite_graph_analytics/pykeen_ops.py
```diff
@@ -0,0 +1,331 @@
...
+        entity_embeddings = actual_model.entity_representations[0]().detach().cpu()
+
+    if entity_embeddings is None:
+        raise AttributeError(
+            f"Cannot extract entity embeddings from model type: {type(actual_model)}. "
+            f"Available attributes: {[attr for attr in dir(actual_model) if not attr.startswith('_')]}"
+        )
+
+    bundle.other[node_embedding_name] = entity_embeddings
+    bundle.other["entity_to_index"] = {entity: idx for idx, entity in enumerate(entity_labels)}
```
darabos, line 313: Instead of dicts, put these in a DataFrame. If we already have (had) a `nodes` table, these could be inserted into that as new columns.

lynxkite-graph-analytics/src/lynxkite_graph_analytics/pykeen_ops.py
```diff
@@ -0,0 +1,331 @@
...
+
+    return bundle
+
+
+@op("Extract embeddings from PyKEEN model")
+def extract_from_pykeen(bundle: core.Bundle, *, node_embedding_name: str, edge_embedding_name: str):
+    bundle = bundle.copy()
+    model = bundle.other["model"]
+    triples = TriplesFactory.from_labeled_triples(
+        bundle.dfs["triples_train"][["head", "relation", "tail"]].values,
```
darabos, line 290: Take table name as a parameter.

Must this be the same table as the one used for the training? If it can only be used with that, then perhaps it's best to extract these at the end of training, as part of the training box.

examples/pykeen-testing.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+@op("Load Embedding into PyKEEN model")
+def load_pykeen_embeddings(bundle: core.Bundle, *, model: str):
+    """
+    Assuming embeddings are torch tensors
+    Entity names has to be a file with comma seperated entity names
+    """
+    node_embeddings = bundle.other["node_embedding"]
+    edge_embeddings = bundle.other["edge_embedding"]
```
darabos, line 80: Stuff that is per-node or per-edge should be stored in the nodes/edges DataFrames.

lynxkite-graph-analytics/src/lynxkite_graph_analytics/pykeen_ops.py
```diff
@@ -0,0 +1,331 @@
...
+    bundle.dfs["training"] = pd.DataFrame({"training_loss": result.losses})
+    bundle.other["model"] = result.model
+
+    return bundle
+
+
+@op("Triples prediction")
+def triple_predict(bundle: core.Bundle, *, table_name: str = "triples_val"):
+    bundle = bundle.copy()
+    model = bundle.other.get("model")
```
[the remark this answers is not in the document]
Mackofi: I think I have done it correctly, renamed the `ModelName` into `PyTorchModelName` and the type into `pytorch-model`, renamed the symbol, so it should've replaced everything using the old naming system. 
Also created a `PyKEENModelName` and `PyKEENModelWrapper` that implements this functionality, and it works flawlessly.

lynxkite-graph-analytics/src/lynxkite_graph_analytics/pykeen_ops.py
```diff
@@ -0,0 +1,331 @@
...
+            create_inverse_triples=req_inverse_triples(model),
+        ),
+    )
+    bundle.dfs["pred"] = pred_annotated.df
+
+    return bundle
+
+
+@op("Full prediction", slow=True)
+def full_predict(bundle: core.Bundle, *, k: int = None):
```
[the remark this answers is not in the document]
Mackofi: This has been fixed in #44.

lynxkite-graph-analytics/src/lynxkite_graph_analytics/pykeen_ops.py
darabos: > Should we convert node/edge embeddings into dataframes from tensors, just so we can then later convert them back into tensors for machine learning use?

Yes.

- Often we can't keep data in tensors because it's too much. Usually neural networks process data in batches sized so that each batch fills up the GPU. The rest of the data is kept on disk or in system memory.
- Tensors often have gradients attached to them which we probably don't want to store.
- Putting the tensors into the DataFrame shouldn't be inefficient. LynxKite uses `cudf.pandas` if `NX_CUGRAPH_AUTOCONFIG` is `true`. In that case I would hope that somehow the various libraries involved in this do a good job. If they don't, that will have to be fixed, but I don't think it should be a design constraint on us.
- This is needed to be compatible with the rest of LynxKite. Per-node/per-edge data is stored in the DataFrames.

> Also I have done the following with `entity_to_index`: I created a `nodes` table during the importing of datasets and then saved the `label`s and `id`s into two columns. Same with relations/edges. But there I saved `source, relation, target, relation_id`.

Awesome!

Mackofi: Alright, I understand, it was just a bit weird to me, but with this I fully understand the reason. Implemented it!

lynxkite-graph-analytics/src/lynxkite_graph_analytics/pykeen_ops.py
```diff
@@ -0,0 +1,331 @@
...
+    # Validation -----------
+    triples = ds.validation.mapped_triples
+    df_val = mapped_triples_to_df(
+        triples=triples,
+        entity_to_id=ds.entity_to_id,
+        relation_to_id=ds.relation_to_id,
+    )
+
+    bundle = core.Bundle()
+    bundle.dfs["triples_train"] = df_train
```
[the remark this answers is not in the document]
Mackofi: I've looked into this further, and found the following:
- Datasets do not have entity properties, so we don't need to worry about that.
- The "Visualize graph" box works quite well with the current `nodes` and `edges` tables, but the edges have to be transformed from their labels into their ids first.
- Aren't relations in a bundle in the context of databases? How would that work with edge types?
- I have not looked into relation selections but filtering is more of a general task in my opinion and should be done in a separate box/SQL query, but if we need to do it in the "Visualize graph" box then I would be able to do that.
- And the last point should already work (with the `edges_train`, `edges_test`, `edges_val` tables), if transformed from their labels to ids, and if we could select which tables are the nodes and edges located at before graph visualization.

lynxkite-graph-analytics/src/lynxkite_graph_analytics/pykeen_ops.py
```diff
@@ -0,0 +1,633 @@
...
+        optimizer=optimizer_type,
+        evaluator=None,
+        random_seed=42,
+        stopper="early",
+        stopper_kwargs=dict(
+            frequency=2,
+            patience=6,
+            relative_delta=0.001,
+            metric="hits@k",
+        ),
```
Mackofi, line 380: Right now there are hard coded, but it would be cool if I (or someone else) came up with a solution to this that does not include just adding 4-5 parameters to the function call...

lynxkite-graph-analytics/src/lynxkite_graph_analytics/pykeen_ops.py
```diff
@@ -163,18 +164,34 @@ class PyKEENModel(str, enum.Enum):
     TransR = "TransR"
     TuckER = "TuckER"
 
-    def to_class(self, triples_factory=None, embedding_dim=None):
+    def to_class(self, triples_factory: TriplesFactory, embedding_dim: int) -> models.Model:
         return getattr(models, self.value)(
             triples_factory=triples_factory,
             embedding_dim=embedding_dim,
+            random_seed=42,
```
darabos, line 171: Let's make the seed controllable by the user! Just make it a parameter of the box. I love having seed parameters — this makes it easy to re-run or not re-run the box.

lynxkite-graph-analytics/src/lynxkite_graph_analytics/pykeen_ops.py
```diff
@@ -192,8 +209,44 @@ def __getattr__(self, name):
     def __str__(self):
         return str(self.model)
 
+    def __getstate__(self):
+        state = {
+            "trained": getattr(self, "trained", False),
```
darabos, line 216: Why not just

```suggestion
            "trained": self.trained,
```

lynxkite-graph-analytics/src/lynxkite_graph_analytics/pykeen_ops.py
```diff
@@ -192,8 +209,44 @@ def __getattr__(self, name):
     def __str__(self):
         return str(self.model)
 
+    def __getstate__(self):
+        state = {
```
darabos, line 215: Instead of listing everything, maybe consider:

```
state = dict(self.__dict__)
del state["model"]
```

lynxkite-graph-analytics/src/lynxkite_graph_analytics/core.py
```diff
@@ -42,17 +42,28 @@
...
         "format": "dropdown",
-        "metadata_query": "[].other.*[] | [?type == 'model'].key",
+        "metadata_query": "[].other.*[] | [?type == 'pytorch-model'].key",
     },
 ]
 """A type annotation to be used for parameters of an operation. ModelName is
 rendered as a dropdown in the frontend, listing the models in the Bundle.
 The model name is passed to the operation as a string."""
 
+PyKEENModelName = typing.Annotated[
```
darabos, line 56: Please move this to the PyKEEN code file. (You could also move PyTorchModelName to the PyTorch code.)

lynxkite-graph-analytics/src/lynxkite_graph_analytics/lynxkite_ops.py
```diff
@@ -135,9 +135,9 @@ def import_graphml(*, filename: str):
 
 @op("Graph from OSM", slow=True)
 def import_osm(*, location: str):
-    import osmnx as ox
+    from osmnx import graph
 
-    return ox.graph.graph_from_place(location, network_type="drive")
+    return graph.graph_from_place(location, network_type="drive")
```
darabos, line 140: Revert this?

lynxkite-graph-analytics/src/lynxkite_graph_analytics/pykeen_ops.py
```diff
@@ -0,0 +1,633 @@
...
+
+    metrics = metrics_str.split(",")
+    metrics_df = pd.DataFrame(columns=["metric", "score"])
+
+    for metric in metrics:
+        metric = metric.strip()
+        try:
+            score = evaluated.get_metric(metric)
+        except Exception as e:
+            raise Exception(f"Possibly unknown metric: {metric}\nError: {e}")
```
darabos, line 626: Use [exception chaining](https://docs.python.org/3/tutorial/errors.html#exception-chaining):

```suggestion
            raise Exception(f"Possibly unknown metric: {metric}") from e
```

darabos (approved):
Awesome, thanks!

lynxkite-graph-analytics/src/lynxkite_graph_analytics/pykeen_ops.py
Mackofi: I think that was co-pilot tab completion just seeing getattr in a lot of places and deciding to use that here too. I have replaced this whole dict with your suggestion of
```python
state = dict(self.__dict__)  
del state["model"] 
```

lynxkite-graph-analytics/src/lynxkite_graph_analytics/pykeen_ops.py
Mackofi: Done, I did it so that the training run uses the same seed as the model.

PR #43, opened 2025-07-11
Type selection on model input parameters

JMLizano:
Adds the ability to choose the type to transform to in the model input mapping. Something we discussed in: https://github.com/biggraph/lynxkite-2000-enterprise/pull/220#discussion_r2137748855


lynxkite-graph-analytics/src/lynxkite_graph_analytics/pytorch/pytorch_core.py
```diff
@@ -66,9 +67,28 @@ def for_sequential(self):
...
+    double = torch.double
+    int = torch.int
+    long = torch.long
+    boolean = torch.bool
+
+
 class ColumnSpec(pydantic.BaseModel):
     df: str
     column: str
+    type: TorchTypes = TorchTypes.float
```
darabos, line 81: Is the binding the right place for this? Can you have a model trained on one type and then run inference with a different type?

I thought the answer is no. In which case this type information belongs inside the model. E.g., as a dropdown in the "Input: tensor" box. So the model would declare that this input is a float/int/bool. And then we can do whatever conversions are necessary.

darabos (commented):
Thanks!

lynxkite-graph-analytics/src/lynxkite_graph_analytics/pytorch/pytorch_core.py
JMLizano: Oh right, that's a very good point, let me change it.

lynxkite-graph-analytics/src/lynxkite_graph_analytics/pytorch/pytorch_core.py
```diff
@@ -377,6 +377,26 @@ def build_model(self) -> ModelConfig:
             self.run_node(node_id)
         return self.get_config()
 
+    def _add_type_information_to_input(self, inputs: list[str]) -> dict[str, dict[str, str | None]]:
```
JMLizano, line 380: I don't like this piece of code too much, this could be simplified a lot if the inputs of a model are limited to input boxes, but I don't know if that's true (can't think of a case it is not though).

Also, I have thought quite a bit about the data structure for `model_inputs`, I went with a dictionary because it seemed the most intuitive/extensible, but let me know if you have other ideas.

lynxkite-graph-analytics/src/lynxkite_graph_analytics/pytorch/pytorch_core.py
```diff
@@ -95,7 +95,7 @@ def _torch_load(data: bytes) -> typing.Any:
 @dataclasses.dataclass
 class ModelConfig:
     model: torch.nn.Module
-    model_inputs: list[str]
+    model_inputs: dict[str, dict[str, str]]
```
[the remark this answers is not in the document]
darabos: > Yeah I gave a bit of though to the type, like having a list of objects (dicts/class) instead of a dict, nothing convinced me, so I went with this.
> 
> I also didn't like having separate fields for each attribute, like the `input_output_types` that you mention . But now that I understand that `model_inputs` != `<input boxes>` I see why you went with that approach. Let me make this change.

Yeah, the downside of the separate fields is that they could become inconsistent. A nice class type for representing all properties would look much better. But the separate fields are so nice for building! We find the different properties of inputs at different stages of the code. It's just easier to collect them separately.

PR #45, opened 2025-07-15
Add NodeSearch improvements

banhidiboti2:


lynxkite-app/web/src/workspace/NodeSearch.tsx
```diff
@@ -12,82 +17,388 @@ export type OpsOp = {
...
+    padding: "6px",
+    boxSizing: "border-box",
+    borderRadius: "4px",
+    border: "1px solid #ccc",
+  },
+  // NEW: Added styled back button for better navigation UX
+  backButton: {
+    marginBottom: "10px",
+    cursor: "pointer",
+    color: "#007bff",
```
darabos, line 391: Let's not introduce new colors either. I suggest you try to work with the existing rules in `index.css`, modifying them and expanding them as needed.

lynxkite-app/web/src/workspace/Workspace.tsx
```diff
@@ -44,7 +44,12 @@ import { usePath } from "../common.ts";
 import EnvironmentSelector from "./EnvironmentSelector";
 import LynxKiteEdge from "./LynxKiteEdge.tsx";
 import { LynxKiteState } from "./LynxKiteState";
-import NodeSearch, { type OpsOp, type Catalog, type Catalogs } from "./NodeSearch.tsx";
+import NodeSearch, {
+  type OpsOp,
+  type Catalogs,
```
darabos, line 49: Unused, right?

lynxkite-app/web/src/workspace/Workspace.tsx
```diff
@@ -247,7 +252,7 @@ function LynxKiteFlow() {
         event.preventDefault();
         setNodeSearchSettings({
           pos: getBestPosition(),
-          boxes: catalog.data![state.workspace.env!],
+          categoryHierarchy: buildCategoryHierarchy(catalog.data![state.workspace.env!]),
```
darabos, line 255: You're still re-building the hierarchy each time the search is opened. Build it when we receive it from the backend.

lynxkite-app/web/tests/lynxkite.ts
```diff
@@ -150,7 +150,7 @@ export class Workspace {
       try {
         await this.tryToConnectBoxes(sourceId, targetId);
         return;
-      } catch (e) {}
+      } catch (e) { }
```
darabos, line 153: I wonder why we get these changes? I would expect Biome to make the formatting uniform. Have you set up pre-commits? (`uvx pre-commit install`)

darabos (commented):
Thanks for the updates! Sorry for the slow review.

lynxkite-app/web/tests/lynxkite.ts
banhidiboti2: Thanks for the help!

darabos (commented):
I've implemented my comments and made some small changes. (Like using backspace for the back button instead of escape, scrolling to the selected item when navigating by keyboard, not scrolling away the search box, selecting the category we came back from, etc.) I think it looks and works great now! Thanks!

PR #55, opened 2025-07-22
lynxkite.core → lynxkite_core & lynxkite → lynxkite-app

darabos:
The package name != module name discrepancy is causing issues when using these packages. I think it's simplest to just rename them. It's early enough that it won't have much effect.

This "frees up" the `lynxkite` package name. I'm thinking it could be a "default" LynxKite installation: basically `lynxkite-app` + `lynxkite-graph-analytics`. So you can just `pip install lynxkite` and get started.


darabos:
Hoping for the best 🎶

PR #62, opened 2025-08-18
Implementations of PyKEEN demos

### LINKED ISSUE #56: PyKEEN: Create demo that uses node embeddings
The primary feature of PyKEEN is making node embeddings that allow good reconstruction of the graph structure. Make a demo where we use such embeddings in a downstream task.

- Pick a dataset. Let's say it's a social network with age/sex/location data.
- Pick a task. Let's say we want to predict if two people are married.
- Use PyKEEN to create node embeddings from the graph structure.
- Train an MLP for which the input is a concatenation of:
  - age/sex/location of person 1 encoded as a vector
  - PyKEEN embedding of person 1 encoded as a vector
  - age/sex/location of person 2 encoded as a vector
  - PyKEEN embedding of person 2 encoded as a vector

  And the output is the probability of being married. (For which we would need to have training data.)

  (Naturally the order of person 1 / person 2 should not matter. You could have the model process `p1` and `p2` separately for a few layers and then sum them: `y = MLPb(MLPa(p1) + MLPa(p2))`. The role of `MLPa` would be to process the `p1`/`p2` vectors so that addition preserves whatever we need to know about them.)

- Measure the performance of this model when trained on the full dataset vs when the age/sex/location data is corrupted vs when the edge data is corrupted. (By corrupted I mean we take a random permutation. So it has the same distribution, but it has zero predictive power.) To be safe, you can also measure the performance when both are corrupted, where we expect to see the performance of random guesses.

The downstream MLP model should use LynxKite's PyTorch model tools and this workspace should show how you can simply transfer embeddings from the PyKEEN boxes to the PyTorch boxes.

That said, maybe implement this in a notebook first (without LynxKite), so we are not chasing LynxKite bugs and data science bugs at the same time. If you can find a similar example that is already implemented then even better.

### LINKED ISSUE #57: PyKEEN: Create link prediction demo
My understand of PyKEEN is that it tries to "compress" the network information into the node embeddings, so that the edges can be reconstructed from them. If such a model is 100% accurate, than it will exactly reproduce the original graph. But if it's less accurate then that, then we can look at the errors and consider them predictions:
- A false positive is an edge the model thinks is likely, but we don't have it in the data. Perhaps it's a prediction for the future?
- A false negative is an edge in the data that the model thought was unlikely. Perhaps it's an anomaly?

In general I don't think we have a way of telling apart errors from useful predictions. But we could find a temporal graph, where edges have a creation timestamp. Split the graph into past (training data) and future (test data). (Maybe PyKEEN includes such a dataset.) Show the accuracy of different models on this dataset.

### LINKED ISSUE #58: PyKEEN: Link prediction demo in inductive setting
@Mackofi has found a good explanation of this use case at https://pykeen.readthedocs.io/en/stable/tutorial/inductive_lp.html#inductive-link-prediction. Let's have a demo of this as well once the simpler cases are done. (#56 #57)

### LINKED ISSUE #59: PyKEEN: Demo that uses node attributes for link prediction
Predicting marriages, the running example in #56, is a link prediction case. But it's clear that the age/sex/location data is very useful for this task. Can the PyKEEN link prediction mechanism use such data? If it's possible, compare against the earlier results in #56.

Mackofi:
I have merged the 4 different demos (#56, #57, #58, #59) into one single pull request, as there were a lot of similarities between demos that had implemented the same feature, and I wanted to check that the demos do work next to each other. But you can still find their own branches if you only want to test one demo at a time.

This is still in a draft state as I am not confident in how the `numeric attributes` and the `MLP` demo reads. Those will be updated, feedback on those and on the more completed `Link Prediction` and `Inductive link prediction` demos are very much welcome.

(Closes #56, Closes #57, Closes #58 and Closes #59)


lynxkite-graph-analytics/src/lynxkite_graph_analytics/pytorch/pytorch_ops.py
```diff
@@ -139,7 +139,7 @@ def layernorm(x, *, normalized_shape=""):
     return torch.nn.LayerNorm(normalized_shape)
 
 
-@op("Dropout", outputs=["outputs", "weights"])
+@op("Dropout")
```
Mackofi, line 142: I removed the outputs from this operator as when I was batching it changed the output shape from `[batch_size, dim]` to `[dim]` and when I was not batching it also did not work.

The wiring was the following: Linear(64) -> Activation(ReLu) -> Dropout(0.2) -> Linear(1) -> Binary Crossentropy with Logits Loss.

When batching the loss could not be calculated as the following error popped up: `ValueError: Target size (torch.Size([607, 1])) must be the same as input size (torch.Size([1]))`

When `batch_size=1`: the `Linear(1)` did not run with the error: `IndexError: Dimension specified as -1 but tensor has no dimensions`

This error also possibly affects the `LayerNorm` box, but I have not tested it yet.

lynxkite-graph-analytics/src/lynxkite_graph_analytics/pykeen_ops.py
```diff
@@ -141,50 +137,178 @@ def import_pykeen_dataset_path(*, dataset: PyKEENDataset = PyKEENDataset.Nations
...
+                        bias=True,
+                        activation=nn.Tanh,
+                    ),
+                )
+            # case "Concat":
+            #     return combination.ConcatCombination(
+            #         dim=int(kwargs.get("dim"))
+            #     )
+            # case "ConcatAggregation":
+            #     return combination.ConcatAggregationCombination('sum', aggregation_kwargs=dict(index=torch.tensor([0,1,2,3,4,1,2,3,4])))
```
darabos: Delete or uncomment these?

lynxkite-graph-analytics/src/lynxkite_graph_analytics/pykeen_ops.py
```diff
@@ -287,36 +496,233 @@ def __str__(self):
...
+def get_inductive_model(
+    bundle: core.Bundle,
+    *,
+    triples_table: core.TableName,
+    inference_table: core.TableName,
+    interaction: PyKEENModel1D = PyKEENModel1D.DistMult,
+    embedding_dim: int = 200,
+    loss_function: str,
+    num_tokens: int = 2,
+    aggregation: str = "mlp",
```
darabos: Make an `enum` for this too, please. It doesn't need to include all PyTorch functions, just what is reasonable to use.

lynxkite-graph-analytics/src/lynxkite_graph_analytics/pykeen_ops.py
```diff
@@ -339,102 +774,242 @@ def train_embedding_model(
...
+
+@op("View early stopping metric", view="visualization")
+def view_early_stopping(bundle: core.Bundle):
+    metric = bundle.dfs["early_stopper_metric"].early_stopper_metric.tolist()
+    v = {
+        "title": {"text": "Early Stopping Metric"},
+        "xAxis": {"type": "category"},
+        "yAxis": {"type": "value"},
+        "series": [{"data": metric, "type": "line"}],
+    }
```
darabos: Oh that's nice! First time I see a really _simple_ visualization with ECharts. 😊

lynxkite-graph-analytics/src/lynxkite_graph_analytics/pykeen_ops.py
Mackofi: I added a few I found useful (sum, mean and the like)

lynxkite-graph-analytics/src/lynxkite_graph_analytics/pykeen_ops.py
```diff
@@ -580,49 +1124,56 @@ class EvaluatorTypes(str, enum.Enum):
...
-        model=model.model,
+        model=model_wrapper.model,
         mapped_triples=testing_triples.mapped_triples,
-        additional_filter_triples=[
-            training_triples.mapped_triples,
-            validation_triples.mapped_triples,
-        ],
-        batch_size=32,
+        additional_filter_triples=additional_filters.mapped_triples,
+        # batch_size=32,
```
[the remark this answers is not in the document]
Mackofi: Added a parameter to customize.

PR #89, opened 2025-11-24
Upgrade all JavaScript dependencies

darabos:
I mainly did this to fix the Dependabot alerts. But it's good to update dependencies anyway. I thought it would break a bunch of things, because it involves a React 18 to 19 upgrade, a Tailwind 3 to 4 upgrade, etc. But it seems fine!

### DIFF
--- a/lynxkite-app/web/src/Tooltip.tsx
+++ b/lynxkite-app/web/src/Tooltip.tsx
@@ -1,21 +1,16 @@
-import { useId } from "react";
+import { renderToStaticMarkup } from "react-dom/server";
 import Markdown from "react-markdown";
-import { Tooltip as ReactTooltip } from "react-tooltip";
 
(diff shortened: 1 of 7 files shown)


darabos:
Haha, it _did_ break everything! Not sure why it seemed to work at first. https://github.com/YousefED/SyncedStore/ looks inactive and it blocked the React 19 switch. Anyway, we're not getting much out of it, because ReactFlow uses copies of the nodes/edges arrays. As long as we are manually copying updates between frontend and backend stores, we can just do that with Y.js. Hence `crdt.ts`. It still needs some work. Got to implement the grouping changes. Moving a box is choppy and emits warnings in the console. Something for next week!

darabos:
I don't see any performance issues now. I'm still a little worried, but I think we really need this. We can't be stuck on old versions. Anyway, let me know if you see any issues!

PR #90, opened 2025-11-28
Box search no longer off screen, zoom on scroll, bottom left buttons removed.

banhidiboti:
Resolves https://github.com/biggraph/lynxkite-2000-enterprise/issues/250 and resolves https://github.com/biggraph/lynxkite-2000-enterprise/issues/251 and resolves https://github.com/biggraph/lynxkite-2000-enterprise/issues/252

### DIFF
--- a/lynxkite-app/web/src/workspace/NodeSearch.tsx
+++ b/lynxkite-app/web/src/workspace/NodeSearch.tsx
@@ -131,16 +131,14 @@ export default function NodeSearch(props: {
   const [selectedIndex, setSelectedIndex] = useState(0);
   const itemRefs = useRef<(HTMLButtonElement | null)[]>([]);
   const searchInputRef = useRef<HTMLInputElement>(null);
-  const containerRef = useRef<HTMLDivElement>(null);
 
   // Calculate adjusted position to keep the component visible
-  const adjustedPos = useMemo(() => {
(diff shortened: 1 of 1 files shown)


lynxkite-app/web/src/workspace/NodeSearch.tsx
```diff
@@ -131,6 +131,35 @@ export default function NodeSearch(props: {
   const [selectedIndex, setSelectedIndex] = useState(0);
   const itemRefs = useRef<(HTMLButtonElement | null)[]>([]);
   const searchInputRef = useRef<HTMLInputElement>(null);
+  const containerRef = useRef<HTMLDivElement>(null);
```
darabos: Remove this.

lynxkite-app/web/src/workspace/NodeSearch.tsx
```diff
@@ -131,6 +131,35 @@ export default function NodeSearch(props: {
...
+
+  // Calculate adjusted position to keep the component visible
+  const adjustedPos = useMemo(() => {
```
darabos: I'm not sure this needs to be a useMemo. If you just want to encapsulate the logic in a function (a good idea), you could put it in a top-level function outside of the component. I think that would be nice and readable.

PR #93, opened 2026-01-06
Remove tooltip question marks, other tooltip improvements

banhidiboti:
In the current tooltips, there are no question marks, they can be accessed by hovering over the title or the label. There is a 1-second delay between hovering and the tooltip appearing. Tooltips no longer pop up underneath another element, and there was a flickering issue which is now gone.

Also Tooltips now rely on a single, global tooltip system.

Resolves https://github.com/biggraph/lynxkite-2000-enterprise/issues/267

### DIFF
--- a/lynxkite-app/web/src/workspace/nodes/NodeParameter.tsx
+++ b/lynxkite-app/web/src/workspace/nodes/NodeParameter.tsx
@@ -14,13 +14,9 @@ const MODEL_OUTPUT_MAPPING = "lynxkite_graph_analytics.ml_ops.ModelOutputMapping
 function ParamName({ name, doc }: { name: string; doc: string }) {
   return (
     <div className="param-name-row">
-      {doc ? (
-        <Tooltip doc={doc}>
-          <span className="param-name">{name.replace(/_/g, " ")}</span>
-        </Tooltip>
-      ) : (
+      <Tooltip doc={doc}>
         <span className="param-name">{name.replace(/_/g, " ")}</span>
-      )}
+      </Tooltip>
     </div>
   );
 }


lynxkite-app/web/src/workspace/nodes/NodeParameter.tsx
```diff
@@ -14,15 +12,15 @@ const MODEL_INFERENCE_INPUT_MAPPING =
...
-  const help = doc && (
-    <Tooltip doc={doc} width={200}>
-      <Help />
-    </Tooltip>
-  );
   return (
     <div className="param-name-row">
-      <span className="param-name">{name.replace(/_/g, " ")}</span>
-      {help}
+      {doc ? (
```
darabos: This check is not necessary. Just do the `<Tooltip>` always. Inside `Tooltip` we check if `doc` is undefined and just return the internal element in that case.

lynxkite-app/web/src/workspace/nodes/LynxKiteNode.tsx
```diff
@@ -182,28 +182,27 @@ function LynxKiteNodeComponent(props: LynxKiteNodeProps) {
...
+                <span className="title-title">{data.title}</span>
+                {data.error && (
+                  <Tooltip doc={`Error: ${data.error}`}>
+                    <AlertTriangle />
+                  </Tooltip>
+                )}
+                {data.collapsed && (
+                  <Tooltip doc="Click to expand node">
+                    <Dots />
+                  </Tooltip>
```
darabos: I just realized we didn't look at this. Tooltips inside tooltips? That's probably not good, right?

I think one tooltip will be enough for the box header.
- When the box is collapsed, show "click to expand node".
- When there's an error, show the error message.
- Otherwise show the documentation.

I recommend moving this logic outside of the JSX. Just do `const tooltip = data.collapsed ? "Click to expand node" : ...;` and then `<Tooltip doc={tooltip}>` in the JSX.

Thanks!

PR #103, opened 2026-01-09
Show only one tooltip on node titles

banhidiboti:


lynxkite-app/web/src/workspace/nodes/LynxKiteNode.tsx
```diff
@@ -209,6 +209,12 @@ function LynxKiteNodeComponent(props: LynxKiteNodeProps) {
     borderColor,
     boxShadow: `0px 5px 30px 0px ${color.toString()}`,
   };
+  const titleTooltip = data.collapsed
+    ? "Click to expand node"
+    : data.error
+      ? `Error: ${data.error}`
+      : data.meta && data.meta.value && data.meta.value.doc;
```
darabos: Where did you get `meta.value` from? There's no `meta.value` anymore. Anyway, let's just use `summary`, so we're consistent:

```suggestion
  const titleTooltip = data.collapsed ? "Click to expand node" : summary;
```

darabos (commented):
Cool, thank you!

PR #123, opened 2026-01-21
Convert Python code to a workspace, live updates, agent

darabos:
For https://github.com/biggraph/lynxkite-2000-enterprise/issues/284. This is just the beginning, but I'm glad to see it works.

image

Being able to look at regular Python code as a LynxKite workspace is cool in itself. It's also a step toward using LynxKite as a vibe-coding aid.

I've got to make the plots work and figure out the right way to use it with CRDT. Do we want to allow changes in LynxKite? Would those update the Python code? I think yes and yes. But also if the .py file changes, we want to update the workspace.


darabos:
It's getting there! The boxes that make plots have to be called `plot_*`. The plots in the screenshot are resized and positioned manually. The default size is too small.

image

Editing the workspace currently overwrites the Python code with the JSON... Will need to fix that.

Mackofi (approved):
Great addition!

Maybe we could also look into giving the AI agent the currently available boxes (as they are guaranteed to work) and make it use those whenever possible. This would probably require that we document all boxes so that the AI knows what box does what, plus its inputs/outputs, but would help out the agent that much more in generating working workspaces.

And in the future maybe we could also do something similar to how Google Colab does things. There is a side-panel in every notebook, and you can ask the AI to make changes to the notebook. 
In LynxKite we could give it the boxes and connections, and instruct it to make the workspace do this and that. Then the agent can make changes in that workspace by adding new boxes, rather than coding the workspace entirely from scratch.

PR #141, opened 2026-02-09
Make ODE-GNN example prettier

darabos:
I've added color and icons, added some comments, and fixed a few issues that popped up along the way.

The change in `groupSelection()` fixes a case where in ODE-GNN/Model if I group the GNN boxes their positions went to NaN. I spent a while trying to understand how that happens, but I still don't know. Maybe the position is still a CRDT Map sometimes? The fix works anyway.

The introduction of `BundleTableView` is not strictly necessary, I just thought the code would be more readable with these dataclasses. Then I realized dataclasses can't pass through CRDT, so I added the conversion. I figure it's still worth it?

The optimizer in PyTorch Models used to be green. I changed it to follow the general convention of green = input and I made the optimizer purple to highlight it. I also made the "Define model" box purple to draw attention to it. It fetches an external thing (the model) without any visual indication. Plus the workspace didn't have enough color! 😄

### DIFF
--- a/examples/ODE-GNN/Experiment.lynxkite.json
+++ b/examples/ODE-GNN/Experiment.lynxkite.json
@@ -16607,10 +16607,11 @@
     },
     {
       "data": {
+        "__execution_delay": null,
         "collapsed": true,
         "display": null,
         "error": null,
-        "expanded_height": null,
+        "expanded_height": 200.0,
         "input_metadata": [
           {
             "dataframes": {
@@ -17085,7 +17086,7 @@
         "title": "Index genes"
       },
       "dragHandle": ".drag-handle",
-      "height": 71.0,
+      "height": 72.0,
       "id": "Index genes 1",
       "position": {
         "x": -3714.30258968075,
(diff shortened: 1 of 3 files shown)


darabos:
> Maybe we could look into supporting more flexible comment boxes? Support for basic HTML and some css possibly?

I had half a heart to do that. There's a relatively easy way: https://github.com/rehypejs/rehype-raw. But it could be a security headache. I hope we can make workspaces as safe as Jupyter notebooks: you can open an untrusted notebook and as long as you don't execute it it can't hurt you.

I think for now we can just add directives for whatever we need. If it gets out of hand we can look at alternatives.

Mackofi (approved):
This example is starting to look nicer and nicer!

The comments of the workspaces seem to be very high-level to me, and skip over a lot of possibly crucial information. I think we could definitely include more things as that would make it even better as having workspaces that basically speaks for themselves is very cool.
But I can also understand how having a lot of text on-screen might be jarring.

PR #147, opened 2026-02-13
Boxes can now display helpful messages from the backend.

Mackofi:
Adds new functionality to the backend, and now we can send helpful messages to be displayed inside the boxes like so:
```python
@op("Sending hello message")
def hello_msg(self, *, name: str):
    self.print(f"Hello {name}!", append=True)
```
Currently has two modes: `replace, append` (using the `append: bool` param). Replace will only show the last message on the UI, while append will keep the message before and append itself onto the end.

Also implements basic functionality covering `tqdm.tqdm` passing the progress bar tqdm gives to the frontend using the above message function. Used like this:
```python
from tqdm import tqdm
@op("Progress bar")
def progress_bar(self, *, name: str):
    with self.tqdm:
        for i in tqdm(range(10)):
            pass
```

The messages currently do not persist inside `@op(slow=True)` boxes, any help fixing that is appreciated!

(Implements https://github.com/biggraph/lynxkite-2000-enterprise/issues/170)

### DIFF
--- a/lynxkite-core/src/lynxkite_core/executors/one_by_one.py
+++ b/lynxkite-core/src/lynxkite_core/executors/one_by_one.py
@@ -2,7 +2,6 @@
 A LynxKite executor that assumes most operations operate on their input one by one.
 """
 
-import asyncio
 import contextlib
 import urllib.parse
 from .. import ops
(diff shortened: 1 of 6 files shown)


darabos (approved):
Awesome!!! Now I love everything about it except the `_cached_with_ctx()` part. But even with that I understand why it's like that and it's probably fine. Thanks a lot for making this happen! 🎉

PR #149, opened 2026-02-16
Add output_metadata and display it on the UI

### LINKED ISSUE #121: Display output count
`select count(1) from nodes` and then a "View tables" is a lot of work for the basic question of "what did we get". Let's try to make this information easily available on the UI, at least in the case where it's cheaply available on the backend. We could do it for single DataFrames and for all tables in `Bundle.dfs`. The table names can make this read really nicely: `123 nodes, 234 edges` if the two tables are `nodes` and `edges`. Let's try displaying this by the output connector.

darabos:
Resolves #121. With multiple tables it gets crowded and we will have to do something about that. But I love it already!

image

image


doragera (approved):
I like it a lot! If there are many labels on the arrow, we could also break it into 2 lines..

darabos:
So, wild story! I forgot to add `output_metadata` in `clean_input()`, so every workspace run triggered another workspace run and so on. This wasn't apparent in manual testing, but slowed things down and caused random failures in the frontend tests. So I kept fixing things in frontend tests.
- I've revived the nested directory tests.
- I've put all test workspaces in a `testing sandbox` folder.
- I've suppressed the delete confirmation popup in tests.
- I've moved the delete command into a dropdown menu.
- I've added renaming in the dropdown menu. And added tests for it.
- Since I added a modal dialog for renaming, I also hooked it up for new workspace/folder creation.
- Other small changes. 😓

darabos:
A thousand apologies if I break something with this messy PR.

PR #171, opened 2026-03-13
Progress page

darabos:
For https://github.com/lynxkite/lynxkite-2000-enterprise/issues/305.

Not connected to the backend, but I think it's useful for discussing the design.

image
image
image

Let's discuss it after GTC!


doragera:
image
image

The progress page is getting connected to the real data from the backend. The user management will be a separate issue, but for now these are the features that are working:
- list active workspaces with progress and buttons for pausing and stopping
- list running k8s deployments with replica count and last workspace that used it, but cannot change replica count from here yet

darabos:
Awesome!

You said you're considering using websockets for keeping the data fresh on these pages. You should also consider CRDT — that's what we use for workspaces. It's not perfect, but it has some advantages. You don't have to worry about dealing with initial load and updates separately. Changes triggered on the frontend are pushed to the backend through the same channel. And it would be the same tech as workspaces, so we don't need to learn two different things. (See `crdt.py` and `crdt.ts`. Because you don't need to integrate with ReactFlow it would be a fair bit simpler for you.)

lynxkite-app/src/lynxkite_app/crdt.py
```diff
@@ -280,14 +280,20 @@ async def workspace_changed(name: str, changes: list[pycrdt.MapEvent], ws_crdt:
...
     ws_simple = ws_pyd.model_copy(deep=True)
     clean_input(ws_simple)
     if ws_simple == last_known_versions.get(name):
         return
     last_known_versions[name] = ws_simple
+    # Only persist for real structural changes. Saving on every execution-state
+    # change (status, telemetry, display) would trigger the file-watcher which
+    # calls load_workspace → crdt_update, overwriting live CRDT state with stale
+    # file data and causing flickering on the progress page.
+    ws_pyd.save(pathlib.Path() / name)
```
darabos: It used to be here, so I know what's wrong with this: this way if you reposition boxes, change comments, etc, the changes are not saved. 😐 

Let's see if we can find a better way. Maybe store the contents somewhere before we write it to the file. And when the watcher sees changes, and the file matches this `last_saved` content then it can ignore the change. It would save some work too.

Or delay saving a little. Do it like `delayed_executions`, so we don't write to the file a thousand times a second. (Or combine this with the first idea?)

The watcher/`crdt_update` have other issues too. I'm making big changes there at the moment. (It doesn't handle node deletion, for example.) It shouldn't impact you because it would be best to avoid running `crdt_update` here altogether.

doragera: I have updated the code with these suggestions. Now we store contents before writing and ignore the changes when the file matches.

doragera:
Resolves https://github.com/lynxkite/lynxkite-2000-enterprise/issues/305 and https://github.com/lynxkite/lynxkite-2000-enterprise/issues/317. Before merge we should consider if some parts of these PR belongs to the enterprise repo.

darabos:
> Resolves [lynxkite/lynxkite-2000-enterprise#305](https://github.com/lynxkite/lynxkite-2000-enterprise/issues/305) and [lynxkite/lynxkite-2000-enterprise#317](https://github.com/lynxkite/lynxkite-2000-enterprise/issues/317). Before merge we should consider if some parts of these PR belongs to the enterprise repo.

Yes, I think this is as enterprise as it gets! Please create a new `lynxkite-enterprise` package in the other repo and move the backend code there. You can follow the example in https://github.com/lynxkite/lynxkite-2000-enterprise/pull/325. (Which I'm going to move to _this_ repo. 😅) Sorry for the extra work. Though I think splitting it to separate files is a good move regardless.

The frontend can stay in this repo. And you can use the same mechanism as in https://github.com/lynxkite/lynxkite-2000/pull/190 — conditionally import `lynxkite_enterprise`, then use `/api/config` to indicate to the frontend that this feature is available, in which case the frontend can display a link to it on the main page.

Thank you!

lynxkite-app/src/lynxkite_app/main.py
```diff
@@ -23,6 +25,12 @@
 except ImportError:
     assistant_router = None
 
+enterprise_backend: Any = None
+try:
+    enterprise_backend = cast(Any, importlib.import_module("lynxkite_enterprise.backend"))
```
darabos: How is the `import_module()` solution better than the simple `import lynxkite_assistant` above? Could we use the same for both optional modules?

lynxkite-app/src/lynxkite_app/crdt.py
```diff
@@ -317,13 +327,23 @@ async def lifespan(app):
     code_websocket_server = CodeWebsocketServer(auto_clean_rooms=False)
     async with ws_websocket_server:
         async with code_websocket_server:
-            yield
+            if enterprise_backend is not None:
+                async with enterprise_backend.lifespan_context(ws_websocket_server):
+                    yield
+            else:
+                yield
```
darabos, line 331: Let's hope we won't have a _third_ module that wants to set up a lifespan thing...

lynxkite-app/src/lynxkite_app/crdt.py
```diff
@@ -317,13 +327,23 @@ async def lifespan(app):
...
     print("closing websocket server")
 
 
 def delete_room(name: str):
     if name in ws_websocket_server.rooms:
         del ws_websocket_server.rooms[name]
+    if name in delayed_executions:
+        delayed_executions[name].cancel()
+        del delayed_executions[name]
+    last_known_versions.pop(name, None)
```
darabos, line 341: Nice.

darabos (commented):
LGTM! Thanks!

I can't approve it, because nominally I'm the author.

lynxkite-app/src/lynxkite_app/main.py
doragera: I changed it to `import`, but I had to add a `# ty: ignore[unresolved-import]`

darabos: > I changed it to `import`, but I had to add a `# ty: ignore[unresolved-import]`

Oh right. That's great! Thanks!

PR #175, opened 2026-03-24
Grouping fixes and selection and dimension consistency

Mackofi:
I was messing around with grouping and ungrouping boxes, and had the same workspace open on two separate windows (to simulate collaborative working) and noticed that when grouping and ungrouping boxes the selections would show on every client. Another issue that I saw was that the box width was not being properly communicated, and would stay fixed on other clients (locally it worked fine).

This PR localizes selections to each client, calculates dimensions properly, and also fixes a long-standing bug where we dealt with `extent` and `parentId` improperly (#81). Now they are correctly deleted when the group ceases to exist.

This is in preparation for the undo/redo feature as undoing a group creation would lead to the console being spammed with errors. (Mainly because of `extent` and `parentId` being set incorrectly)

### DIFF
--- a/lynxkite-app/web/src/workspace/Workspace.tsx
+++ b/lynxkite-app/web/src/workspace/Workspace.tsx
@@ -418,7 +418,6 @@ function LynxKiteFlow() {
       width: 0,
       height: 0,
       data: { title: "Group", params: {} },
-      selected: true,
     };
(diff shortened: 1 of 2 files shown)


darabos (approved):
Thanks! I hope it's the last bug in `crdt.ts`! 🤞

PR #176, opened 2026-03-31
Implements undo functionality

### LINKED ISSUE #115: Undo workspace edits
We're using CRDT. It should be possible to undo changes.

Mackofi:
This PR adds the ability to undo/redo actions (closes #115). The user is able to undo/redo their own changes, but only during the same session. This is probably the expected behavior we want.

We don't explicitly make use of the stored `crdt`, but with using [`Y.UndoManager`](https://docs.yjs.dev/api/undo-manager) we can get away with very little code changes and we get a lot of convenience. For example: we don't need to track what users did what, the `UndoManager` tracks it for us.

### DIFF
--- a/lynxkite-app/web/src/workspace/Workspace.tsx
+++ b/lynxkite-app/web/src/workspace/Workspace.tsx
@@ -527,6 +527,7 @@ function LynxKiteFlow() {
                   className="btn btn-link"
                   disabled={selected.length < 2}
                   onClick={groupSelection}
+                  name="groupBtn"
                 >
                   <GroupIcon />
                 </button>
@@ -536,6 +537,7 @@ function LynxKiteFlow() {
                   className="btn btn-link"
                   disabled={!isAnyGroupSelected}
                   onClick={ungroupSelection}
+                  name="ungroupBtn"
                 >
                   <UngroupIcon />
                 </button>
--- a/lynxkite-app/web/src/workspace/Workspace.tsx
+++ b/lynxkite-app/web/src/workspace/Workspace.tsx
@@ -169,16 +169,23 @@ function LynxKiteFlow() {
       } else if (event.key === "r") {
         event.preventDefault();
         executeWorkspace();
-      } else if (isPrimaryModifierPressed && !(nodeSearchSettings || isTypingInFormElement())) {
-        if (event.key.toLowerCase() === "c") {
-          copySelection(nodes, edges, crdt?.ws ?? {});
-        } else if (event.key.toLowerCase() === "v") {
-          pasteSelection(crdt, cursorScreenPos, reactFlow, setMessage);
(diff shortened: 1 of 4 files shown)


lynxkite-app/web/src/workspace/Workspace.tsx
```diff
@@ -169,16 +169,23 @@ function LynxKiteFlow() {
...
-          cutSelection(nodes, edges, crdt?.ws ?? {}, deleteSelection);
-        } else if (event.key.toLowerCase() === "a") {
-          event.preventDefault();
-          selectAll();
+      } else if (isPrimaryModifierPressed) {
+        if (event.key === "z") {
+          crdt?.undo();
+        } else if (event.key === "y") {
+          crdt?.redo();
+        } else if (!(nodeSearchSettings || isTypingInFormElement())) {
```
darabos, line 177: Are you sure about this? I think if I press Ctrl-Z in a textbox, I probably want to undo the editing inside the textbox, not the last change in the workspace. It could be lumped in with the copy/paste keys.

lynxkite-app/web/tests/undo_redo.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+});
+
+test("undo/redo normal text input", async () => {
+  await workspace.addBox("NetworkX › Generators › Directed › Scale-free graph");
+  const graphBox = workspace.getBox("Scale-free graph 1");
+  await graphBox.getByLabel("n", { exact: true }).fill("10");
+  await workspace.page.keyboard.press("Control+z");
+  await expect(graphBox.getByLabel("n", { exact: true })).toHaveValue("");
+  await workspace.page.keyboard.press("Control+y");
+  await expect(graphBox.getByLabel("n", { exact: true })).toHaveValue("10");
```
darabos, line 111: Huh, okay, if this works then maybe it's better this way. Thanks for the fantastic tests!

darabos (approved):
Way to go!

lynxkite-app/web/src/workspace/Workspace.tsx
Mackofi: As text input also carries `crdt` updates the `UndoManager` should also track those. We also get the added benefit of being able to undo changes after we have clicked out of the text input (or `Comment` box!).

lynxkite-app/web/tests/undo_redo.spec.ts
Mackofi: Yeah, I wanted to make sure most features still work as expected. Any more ideas for what else should we test?

darabos: > Yeah, I wanted to make sure most features still work as expected. Any more ideas for what else should we test?

Nope! Merge away! 🚀

PR #177, opened 2026-03-31
Early stage of the easy connections

### LINKED ISSUE #119: Easy box connection
When you're zoomed out it's pretty hard to connect boxes because the handles are so small. ReactFlow has an example for ["proximity connect"](https://reactflow.dev/examples/nodes/proximity-connect) which is similar to what we did in the old LynxKite. I think our solution was even nicer:
- If you're moving a box and it's close to another box, an edge is added between them.
- This new edge is weak. If you move close to another box (in the same movement) the edge is replaced by a new edge for the second box.
- But moving farther from the box does not break the new edge.
- None of this affects old edges.

banhidiboti:
Resolves #119.


lynxkite-app/web/src/workspace/Workspace.tsx
```diff
@@ -60,6 +60,43 @@ const PlayIcon = memo(Play);
 const PauseIcon = memo(Pause);
 const CloseIcon = memo(Close);
 const ChangeTypeIcon = memo(Transfer);
+const MIN_AUTO_CONNECT_DISTANCE = 220;
+
+function allowsMultipleConnections(inputType: unknown): boolean {
```
darabos: Wow, this is complex! I don't think we can safely make any guesses like this. Different executors may have very different semantics.

How about a simpler solution: Just assume everything allows multiple connections. I hope that will be fine. If it's annoying, we can switch to assuming nothing allows multiple connections.

lynxkite-app/web/src/workspace/Workspace.tsx
```diff
@@ -305,24 +335,245 @@ function LynxKiteFlow() {
...
     [crdt],
   );
-  const parentDir = parentPath(path!);
+
+  function getAbsPos(node: Node) {
+    const internal = reactFlow.getInternalNode(node.id);
+    return internal?.internals.positionAbsolute ?? node.position;
+  }
+
+  function getNodeHandles(node: Node) {
```
darabos: That's a lot of code! Is it the same code as in LynxKiteNode.tsx? If so, could you please de-duplicate it?

lynxkite-app/web/src/workspace/Workspace.tsx
```diff
@@ -305,24 +335,245 @@ function LynxKiteFlow() {
...
+    const width = node.width ?? 200;
+    const height = node.height ?? 200;
+    const offset = (handle.offsetPercentage ?? 50) / 100;
+    if (handle.position === "left") return { x: p.x, y: p.y + height * offset };
+    if (handle.position === "right") return { x: p.x + width, y: p.y + height * offset };
+    if (handle.position === "top") return { x: p.x + width * offset, y: p.y };
+    return { x: p.x + width * offset, y: p.y + height };
+  }
+
+  function findClosestHandlePair(sourceNode: Node, targetNode: Node) {
```
darabos: I think some of these functions could easily be moved into a separate file. Just to make it easier to navigate Workspace.tsx.

lynxkite-app/web/src/workspace/Workspace.tsx
```diff
@@ -305,24 +335,245 @@ function LynxKiteFlow() {
...
+          closeEdge.source,
+          closeEdge.sourceHandle!,
+          closeEdge.target,
+          closeEdge.targetHandle!,
+        )
+      ) {
+        setPreviewEdge(null);
+        return;
+      }
+      setPreviewEdge({
```
darabos: It's possible that a preview edge already exists, and it's the same as `closeEdge`. In this case we don't need to call `setPreviewEdge()`. This may be a meaningful saving, as it avoids triggering a re-render on ever frame when a node is moving. (Although it's possible that already triggers a re-render anyway.)

darabos (commented):
Hi! Thanks! I haven't tried it, but last time you showed it it looked fantastic!

So what's the status of this? Is there still an issue with something? Why is the PR called "early stage"? Is this not ready?

darabos (commented):
I kept promising you I would review this PR. But it looks like I already reviewed it a month ago! Have you seen my comments?

lynxkite-app/web/src/workspace/Workspace.tsx
banhidiboti: Simplified allowsMultipleConnections to always return true, removing all the complex type-guessing logic.

lynxkite-app/web/src/workspace/Workspace.tsx
banhidiboti: Exported getHandles from LynxKiteNode.tsx and replaced the duplicate getNodeHandles body in Workspace.tsx with a call to it (~50 lines removed, replaced with 5)

lynxkite-app/web/src/workspace/Workspace.tsx
banhidiboti: Extracted the auto-connect helpers into a new file named autoConnect.ts

lynxkite-app/web/src/workspace/Workspace.tsx
banhidiboti: Optimized onNodeDrag to skip setPreviewEdge when the preview edge hasn't changed, avoiding unnecessary re-renders during drag.

darabos (approved):
@banhidiboti Please take a look as well!

PR #185, opened 2026-04-09
Compute all kinds of degree at once

darabos:
This is probably the cheapest operation. Don't make me place three boxes to calculate all three degree attributes!


doragera (approved):
Looks good!

PR #187, opened 2026-04-10
Different ways to merge Bundles

### LINKED ISSUE #68: Offer different merge strategies in Organize
The Organize box in Graph Analytics takes multiple bundles and combines them into one bundle. Works great when all the tables have different names. But what do we do when two bundles have a table with the same name?

Currently we just take the "last" one. But I think we should offer different strategies:
- Raise an error. (Could be the default, so the user makes a conscious choice.)
- Rename them. `df` + `df` &rarr; `df 1` + `df 2` I'm not sure how we would tell which one is `1` and is `2`.
- Concatenate them if they have the same schema.
- Concatenate them as graphs. (I'm not sure what this would take beyond concatenating the nodes and edges tables.)
- Just take any one of them. The current behavior. I'm not sure when we would choose this.

darabos:
Edit: resolves https://github.com/lynxkite/lynxkite-2000/issues/68.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,7 +20,7 @@ repos:
   hooks:
     - id: ty-check
       name: ty-check
-      language: python
+      language: system
       entry: uv run --no-sync ty check
       types: [python]
       pass_filenames: true
--- a/lynxkite-graph-analytics/src/lynxkite_graph_analytics/bundle.py
+++ b/lynxkite-graph-analytics/src/lynxkite_graph_analytics/bundle.py
@@ -1,0 +1,313 @@
+"""Bundle data structures and related helpers."""
+
+import dataclasses
+import enum
+import typing
+
+import networkx as nx
+import pandas as pd
+import polars as pl
+
+from .relation_definition import RelationDefinition
+
+
+class BundleMergeMode(enum.StrEnum):
+    """How `Bundle.merge` handles name collisions.
+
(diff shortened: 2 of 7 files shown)


Mackofi:
With this maybe we could also take another look at #68?

darabos:
> With this maybe we could also take another look at #68?

This was such a nice small PR. 😭 

I've been meaning to split Bundle into its own file for a while, as it seemed to be growing. And these merge functions got out of hand. (Largely AI written, but I don't think they could be made simpler.) I still import Bundle and RelationDefinition to `core.py` to avoid breaking too many things at once.

Can you please take a look?

As for the pre-commit change, it's to fix this warning:

```
warning: `VIRTUAL_ENV=/home/darabos/.cache/pre-commit/repozxow9bd9/py_env-python3.12` does
not match the project environment path `.venv` and will be ignored; use `--active` to
target the active environment instead
```

doragera:
Please add some tests for the merge behaviors

darabos:
> Please add some tests for the merge behaviors

Added, thanks! Is it okay otherwise? Or do you think this could be made simpler? We don't need three different modes if nobody's going to use some of them.

Mackofi (approved):
> We don't need three different modes if nobody's going to use some of them.

I could see uses for all three. I think we should keep it and if we don't actually use it we can remove it later.

The old behavior was different from all of these, could break some workspaces, I think we can live with that.
Otherwise looks good to me!

PR #188, opened 2026-04-10
Slashes between posix and windows

### LINKED ISSUE #183: Windows slashes
Lívia has tried using LynxKite on Windows. It starts and workspaces load and many things work. But some parts of LynxKite assume that path separators are `/` and not `\`. I think searching for `"/"` leads us to most of the suspicious bits.

Mackofi:
I tried to find all possible places where paths could cause an issue. I have not found any more so hopefully now path behavior is the same between windows and posix based systems.

(Resolves #183)


lynxkite-app/src/lynxkite_app/crdt.py
```diff
@@ -4,6 +4,7 @@
 import contextlib
 import enum
 import pathlib
+import posixpath
```
darabos, line 6: Is this an undocumented part of the standard library? 👀 Is there no `relpath()` on Windows, or what's the story here?

lynxkite-app/src/build_frontend.py
```diff
@@ -12,8 +12,8 @@ def run(self):
         here = Path(__file__).parent.parent
         frontend_dir = here / "web"
         package_dir = here / "src" / "lynxkite_app" / "web_assets"
-        subprocess.check_call(["npm", "install"], cwd=frontend_dir)
-        subprocess.check_call(["npm", "run", "build"], cwd=frontend_dir)
+        subprocess.check_call("npm install", shell=True, cwd=frontend_dir)
+        subprocess.check_call("npm run build", shell=True, cwd=frontend_dir)
```
darabos: Question 1: Why this change?
Question 2: I wish we could delete this whole script. The error handling is horrible, and it forces everyone to have frontend build tools. Maybe we could build the frontend in a GitHub Action and commit the artifacts into the repo. What do you think?

darabos (commented):
I like the approach of hiding the situation from the frontend! 👍 And most of the backend too. Only need to pay attention where we read/write something.

If a workspace has an Import CSV with a file name param `uploads/data.csv`, will that work? It's a bit scary to think about all the places file IO can surface. Is there no holistic solution?

lynxkite-app/src/lynxkite_app/crdt.py
Mackofi: Maybe undocumented? Not sure, but `posixpath` is basically `os.path` but it always behaves as if we were on a posix filesystem, available on all platforms.

lynxkite-app/src/build_frontend.py
Mackofi: Answer 1: For some unknown reason `npm` on Windows is `npm.CMD`?? (which is a batch file??) 
```
> python -c "import shutil, subprocess; npm = shutil.which('npm'); print('Found:', npm); subprocess.check_call([npm, '--version'])"
Found: C:\Program Files\nodejs\npm.CMD
11.12.1
```
So with `shell=True` I could run batch files, but I might've found a better solution (avoiding `shell=True`):
```python
npm = shutil.which("npm")
assert npm, "npm not found on PATH"
subprocess.check_call([npm, "install"], cwd=frontend_dir)
subprocess.check_call([npm, "run", "build"], cwd=frontend_dir)
```

Answer 2: If we choose the commit the artifacts into the repo we would be possibly committing 36mb (`web/dist` folder) or 61mb (`web_assets` folder) into the repo. We also somehow do save the frontend assets when installing with `pip install lynxkite` (no clue what is happening there). So for running atleast we don't need `npm`, but for development we do.

Mackofi:
>  It's a bit scary to think about all the places file IO can surface. Is there no holistic solution?

Sadly I don't think there is a holistic solution for this. We just have to be careful to use `pathlib.Path` for filesystem and posix strings for identifiers.

> If a workspace has an Import CSV with a file name param `uploads/data.csv`, will that work?

Filename params are server dependent. If the server uses a posix filesystem only `/` is accepted, but on Windows both `\\` and `/` work.

darabos:
> > If a workspace has an Import CSV with a file name param `uploads/data.csv`, will that work?
> 
> Filename params are server dependent. If the server uses a posix filesystem only `/` is accepted, but on Windows both `\\` and `/` work.

Oh, both work on Windows? That's great! Then we don't need to worry too much.

lynxkite-app/src/build_frontend.py
darabos: > Answer 1: For some unknown reason `npm` on Windows is `npm.CMD`?? (which is a batch file??)

Huh, okay, thanks.

> Answer 2: If we choose the commit the artifacts into the repo we would be possibly committing 36mb (`web/dist` folder) or 61mb (`web_assets` folder) into the repo. We also somehow do save the frontend assets when installing with `pip install lynxkite` (no clue what is happening there). So for running atleast we don't need `npm`, but for development we do.

For `pip install lynxkite` it's grabbing the published wheel file which already includes the frontend build. (https://pypi.org/project/lynxkite-app/0.3.0/#files) It's only 8.6 MB compressed. Even so that's too much to include with every PR.

Oh well. Let's keep our eyes open and maybe we find a better solution one day!

lynxkite-app/src/lynxkite_app/crdt.py
darabos: > Maybe undocumented? Not sure, but `posixpath` is basically `os.path` but it always behaves as if we were on a posix filesystem, available on all platforms.

Yeah, I see — because I've looked it up in the Python library source code! How do you even know about this package?

Anyway, I'm okay with it. But please add a comment to give some pointers to future readers. Thanks!

darabos (approved):

PR #190, opened 2026-04-15
CRDT fixes and frontend for assistant

darabos:
A bit of a mixed bag, but this is all that was needed to enable the assistant.


lynxkite-app/src/lynxkite_app/crdt.py
```diff
@@ -353,14 +296,20 @@ async def code_changed(name: str, changes: pycrdt.TextEvent, text: pycrdt.Text):
...
 
 
-def get_room(name):
-    return ws_websocket_server.get_room(name)
+async def get_room(name):
+    return await ws_websocket_server.get_room(name)
+
+
+def get_room_if_exists(name):
+    return ws_websocket_server.rooms.get(name)
```
doragera, line 304: This function name can be misleading. If the room does not exist it returns a None there is no check here. Also no check for None where the function is called.

lynxkite-app/src/lynxkite_app/main.py
```diff
@@ -45,6 +54,11 @@ def get_catalog(workspace: str):
     return {env: _get_ops(env) for env in ops.CATALOGS}
 
 
+@app.get("/api/config")
+def get_config() -> dict[str, bool]:
+    return {"assistant_available": assistant_router is not None}
```
doragera, line 59: `/api/config` is a very general endpoint name, are we going to use it for other configs? Should we track here if the enterprise features are available or not?

doragera (approved):
Looks good!

lynxkite-app/src/lynxkite_app/crdt.py
darabos: > This function name can be misleading. If the room does not exist it returns a None there is no check here.

How about `get_room_or_none()`?

> Also no check for None where the function is called.

Oops! Will fix. We've decided to open-source the backend too, so I will come back with a new pull request and I'll fix it there. Thanks for the comments!

lynxkite-app/src/lynxkite_app/main.py
darabos: > `/api/config` is a very general endpoint name, are we going to use it for other configs? Should we track here if the enterprise features are available or not?

Yes, it's intentionally general. We can put any kind of global information here. (Anything not workspace-specific.) For example, whether user authentication is required or not, etc. But for now, I only have `assistant_available`, so it looks a bit silly.

PR #191, opened 2026-04-16
LynxKite assistant

darabos:
Resolves https://github.com/lynxkite/lynxkite-2000-enterprise/issues/322. Supercedes https://github.com/lynxkite/lynxkite-2000/pull/190.

The backend and frontend happily together! For testing I've used `export LYNXKITE_ASSISTANT_MODEL='openai:gpt-5.4-mini'`.


darabos:
The tests pass! I've addressed the review comments from #190 and https://github.com/lynxkite/lynxkite-2000-enterprise/pull/325. @doragera can I ask you for another round of review on the backend side? (Stuff in `lynxkite-assistant`.) Thank you!

doragera (approved):
I like this assistant a lot. Tried it locally but got an import error:
`ImportError: Initializing ChatOpenAI requires the langchain-openai package. Please install it with pip install langchain-openai`

After installing it worked smoothly, add it to the dependencies please.

darabos:
> I like this assistant a lot. Tried it locally but got an import error: `ImportError: Initializing ChatOpenAI requires the langchain-openai package. Please install it with pip install langchain-openai`
> 
> After installing it worked smoothly, add it to the dependencies please.

Cool, thanks for testing it! I didn't add `langchain-openai` by default because I figured people can install the provider-specific package that is specific to their chosen provider. But it makes things a bit easier if we have it by default, and I don't think it's a big package. Installed! I'll merge if the tests pass.

PR #199, opened 2026-04-24
Add /progress-demo

darabos:


doragera (approved):
I'm still not an expert in frontend, but LGTM!

PR #210, opened 2026-04-30
Add download button to View molecule node

### LINKED ISSUE #207: Allow downloading molecules
We can visualize molecules but not download them. I think it's natural to look at them and then decide whether you want to save it. So how about adding a download feature to the molecule visualization box?

doragera:
image

Feel free to redesign the Download button a bit!

resolves #207

### DIFF
--- a/lynxkite-app/web/src/workspace/nodes/NodeWithMolecule.tsx
+++ b/lynxkite-app/web/src/workspace/nodes/NodeWithMolecule.tsx
@@ -2,13 +2,35 @@ import React, { type CSSProperties, useEffect } from "react";
 import LynxKiteNode from "./LynxKiteNode";
 import { NodeWithParams } from "./NodeWithParams";
 
+type MoleculeDisplayConfig = {
+  data?: string;
+  format?: string;
+  filename?: string;
+  ligand?: string;
+};
+
 const NodeWithMolecule = (props: any) => {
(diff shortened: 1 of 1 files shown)


lynxkite-app/web/src/workspace/nodes/NodeWithMolecule.tsx
```diff
@@ -2,13 +2,35 @@ import React, { type CSSProperties, useEffect } from "react";
...
+    anchor.remove();
+    URL.revokeObjectURL(url);
+  };
+
   useEffect(() => {
-    const config = props.data?.display;
-    if (!config || !containerRef.current) return;
+    const displayConfig = props.data?.display as MoleculeDisplayConfig | undefined;
+    if (!displayConfig || !containerRef.current) return;
+    const config = displayConfig;
```
darabos, line 33: Do we need the `config` / `displayConfig` distinction? Seems like they are the same.

```suggestion
    const config = props.data?.display as MoleculeDisplayConfig | undefined;
    if (!config || !containerRef.current) return;
```

darabos (approved):
Looks great, thanks!!!

darabos:
Sorry! I just discovered Molstar has model downloads too!

Screenshot 2026-05-01 at 8 49 05

Is there anything it cannot do??

Issue #208, opened 2026-04-29
Allow viewing multiple structures in molecule viewer


darabos:
A very powerful feature of PyMOL is that it shows the list of entities and you can toggle which ones you want to view at the same time.

For us this is extra tricky because we have multiple dimensions across which you can get molecules:
- The structure comes from a table which generally has more than one row. These are, for example, different drug candidates.
- You have multiple tables. E.g. with rfDiffusion you have the original protein in one table and the generated binder in another.
- You may connect multiple inputs to the visualization box. (Not currently supported but I think it would be a good change.)

Let's come up with a quick solution for this that at least supports the rfDiffusion case. Then we can iterate on a more complete solution if needed.

PR #212, opened 2026-04-30
Use Mol*

darabos:
For #208. We still need something on the backend if we want to send structures from two different sources to the frontend. But it's amazing what Molstar can do. I'm in love.

https://github.com/user-attachments/assets/64841e9b-5bc4-4296-8068-1eab99b709a3


banhidiboti (approved):
This looks fantastic!

Issue #209, opened 2026-04-29
Allow reading file contents into a 1-row 1-column table


darabos:
We have pretty good tools for tables, and for example we take molecular structures from tables. If you have your molecular structure in a CIF file, it's pretty tricky to get it into a table though. You can't read it as a CSV because it has many rows.

PR #213, opened 2026-05-07
Handle .cif file formats

doragera:
My attempt at #209
Copying large file contents into a box seems a bit inconvenient for me. Supporting .cif formats at drag-and-drop and in "Import file" box is much more intuitive but then what about a .pdb? Should we include that too? And there are also hundreds of other file formats that we can just read into a 1-row 1-column table.

(testing this with the View Molecule box was a headache as I only realized after too much debugging that not every .cif content can be visualized)


darabos (approved):
Yeah I feel like we should one day sit down and think through how we want file imports to work. And that day should probably have been a year ago. 😅 We could discuss it next week (including @Mackofi).

But for now I think this is a great solution! Thank you!

PR #230, opened 2026-05-26
Changed default box size from 200x200px to 300x300px

banhidiboti:


darabos (commented):
Thanks! Please make it a multiple of the grid size.

darabos (approved):
👍  I haven't looked at the tests, but hopefully it's not too bad.

PR #234, opened 2026-05-27
Fixes custom enums not being recognized.

### LINKED ISSUE #206: Caching doesn't work if custom box has an enum parameter
Is this true? This seems to be the case for me, but I'm in the middle of a pretty complex workspace.

`Can't pickle <enum 'TumorTypes'>: it's not found as gnn_to_text.TumorTypes` So this may "only" be an issue for boxes declared in user scripts. Still important to fix, but it may explain why we didn't notice it sooner.

StrEnums are easy to convert back and forth. It's a shame if they are not pickleable.

Mackofi:
This seems way too easy for a fix, but caching now works! 

This also includes collision prevention for two scripts named the same (e.g. `folder1/utils.py` and `folder2/utils.py`), and avoids overriding built-in modules too.

(Fixes #206)


darabos (commented):
Thanks for trying this!

I would have attempted to make a subclass of StrEnum with custom pickling methods, and use that for the parameter enums. But I think that also wouldn't work. 😩 Give up or fork Joblib? Do you have more ideas?

darabos (approved):
This looks nice! Does it work?

PR #238, opened 2026-06-01
Add UserMenu component with auth utilities

banhidiboti:
Resolves https://github.com/lynxkite/lynxkite-2000-enterprise/issues/245


lynxkite-app/web/src/auth.ts
```diff
@@ -0,0 +1,136 @@
+// Keycloak OIDC authentication utilities using Authorization Code flow with PKCE.
```
darabos: Delete this file. Turns out you already have the auth information on the frontend through the `oidc-client-ts` library. Put this in `common.ts`:

```typescript
export function useAuth() {
  const [user, setUser] = useState<User | null>(null);
  useEffect(() => {
    const manager = getUserManager();
    if (!manager) {
      return;
    }
    manager.getUser().then(setUser);
    const onUserLoaded = (u: User) => setUser(u);
    const onUserUnloaded = () => setUser(null);
    manager.events.addUserLoaded(onUserLoaded);
    manager.events.addUserUnloaded(onUserUnloaded);
    return () => {
      manager.events.removeUserLoaded(onUserLoaded);
      manager.events.removeUserUnloaded(onUserUnloaded);
    };
  }, []);
  return user;
}
```

Then anywhere you want you can just say `const user = useAuth();` and use `user?.profile?.name`.

Image

banhidiboti: Done

lynxkite-app/web/src/UserMenu.tsx
```diff
@@ -0,0 +1,63 @@
+import { useState } from "react";
```
darabos: Please add a top-level comment to introduce what this file is about.

lynxkite-app/web/src/UserMenu.tsx
```diff
@@ -0,0 +1,63 @@
...
+import Login from "~icons/tabler/login";
+import Logout from "~icons/tabler/logout";
+import UserCircle from "~icons/tabler/user-circle";
+import UserPlus from "~icons/tabler/user-plus";
```
darabos: At one point the largest impact on our performance was surprisingly due to the icons in the toolbar. (https://github.com/lynxkite/lynxkite-2000/pull/130) These also are icons in the toolbar. Please make sure they don't hurt the performance.

lynxkite-app/web/src/UserMenu.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+export default function UserMenu() {
+  const [open, setOpen] = useState(false);
+  const user = useAuth();
+  const loggedIn = !!user && !user.expired;
+  const userName = user?.profile?.name || "User";
+  const userEmail =
+    user?.profile?.email || user?.profile?.preferred_username || user?.profile?.name;
+
+  if (!loggedIn) {
```
darabos, line 27: If there is no user it generally means there is no authentication. In this case display nothing.

The user will be redirected to the login page if they hit a permission denied error.

lynxkite-app/web/src/UserMenu.tsx
```diff
@@ -0,0 +1,63 @@
...
+        title={userName}
+      >
+        <UserCircle />
+        <span className="user-menu-label">{userName}</span>
+      </button>
+      {open && (
+        <ul className="dropdown-content menu shadow-lg rounded-box bg-base-100 z-50 w-52 p-2 mt-2">
+          <li className="menu-title px-4 py-2 text-sm opacity-70">{userEmail}</li>
+          <li>
+            <button type="button" onClick={() => void triggerLogout()}>
```
darabos, line 58: What does `void` do here?

lynxkite-app/web/src/common.ts
```diff
@@ -73,6 +73,36 @@ export async function triggerLogin() {
   }
 }
 
+export async function triggerRegister() {
```
darabos: Can't we leave registration to Keycloak? In most deployments there will be no registration. LynxKite is typically deployed for internal use. Keycloak integrates with the existing authentication (e.g. AD). Users can easily log in. But we don't handle registration.

Some deployments, like public demo instances, may want registration. I think the natural place for this would be on the Keycloak login page. Users are redirected to the login page when they hit a permission denied error. They can either log in there, or — if enabled in Keycloak — sign up.

darabos (commented):
Thank you!

lynxkite-app/web/src/common.ts
banhidiboti: Agreed. I've removed the triggerRegister function from the frontend.

lynxkite-app/web/src/UserMenu.tsx
banhidiboti: It turns a Promise into undefined, satisfying React's expectation that event handlers return nothing.

lynxkite-app/web/src/UserMenu.tsx
banhidiboti: Done

lynxkite-app/web/src/UserMenu.tsx
banhidiboti: Fixed — wrapped all three icons with React.memo, same pattern as the toolbar icons in Workspace.tsx from #130.

lynxkite-app/web/src/UserMenu.tsx
```diff
@@ -0,0 +1,66 @@
...
+          type="button"
+          className="user-menu-button"
+          onClick={() => void triggerLogin()}
+          title="Sign In"
+        >
+          <Login /> <span className="user-menu-label">Sign In</span>
+        </button>
+      </div>
+    );
+  }
```
darabos: I see that this is pretty, but it generally won't work. When you are not logged in, you won't even be able to list the workspaces. So you won't be able to open a workspace and won't see this "Sign In" button.

At least that's the theory. `list_dir()` in `main.py` calls `auth.check_permission()`, which calls `get_current_user()`, which raises an `HTTPException(status_code=401)` if you're not logged in. Is this not how it works for you?

I think it was fine the way it was. With `triggerLogin()` called on 401 errors in `common.ts`.

banhidiboti: Done

lynxkite-app/web/src/Directory.tsx
```diff
@@ -74,11 +74,10 @@ function Breadcrumbs(props: { path: string }) {
 }
 
 export default function Directory() {
-  useConfig();
```
darabos: Haha, we had this line?

lynxkite-app/web/src/Directory.tsx
```diff
@@ -74,11 +74,10 @@ function Breadcrumbs(props: { path: string }) {
...
   const path = usePath().replace(/^[/]$|^[/]dir$|^[/]dir[/]/, "");
   const encodedPath = encodeURIComponent(path || "");
   const config = useConfig();
-  const list = useSWR(`/api/dir/list?path=${encodedPath}`, fetcher, {
+  const list = useSWR(config.data ? `/api/dir/list?path=${encodedPath}` : null, fetcher, {
```
darabos: This is to delay loading the listing until the config is loaded? Do we need to do that?

lynxkite-app/web/src/UserMenu.tsx
```diff
@@ -0,0 +1,66 @@
...
+// Re-rendering icons is expensive in dev mode; memoizing prevents it.
+const Login = memo(LoginIcon);
+const Logout = memo(LogoutIcon);
+const UserCircle = memo(UserCircleIcon);
+
+export default function UserMenu() {
+  const [open, setOpen] = useState(false);
+  const { data: config } = useConfig();
+  const user = useAuth();
+  const authEnabled = !!config?.authentication_issuer;
```
darabos, line 17: You could `return null` right away if this is false. It would make the branching logic a bit simpler.

lynxkite-app/web/src/common.ts
darabos, line 138: Are these changes needed? And if so, why?

It looks like you're moving the login trigger from `apiFetch()` to `apiJson()`. This is weird. Do we want login redirects for 401s on JSON requests, but not on 401s on other requests?

darabos (commented):
Thank you! Sorry for dragging out this review, but you keep making changes that seem to be outside of the scope of the PR. That's fine, but for these "surprise" changes it's best if you explain them ahead of time. You know I'm going to ask in the code review otherwise. 😅

lynxkite-app/web/src/Directory.tsx
banhidiboti: Yes, we need to delay — if requests fire before config loads, any 401 redirect silently does nothing.

lynxkite-app/web/src/UserMenu.tsx
banhidiboti: Good point, done.

lynxkite-app/web/src/common.ts
banhidiboti: You're right, direct apiFetch() callers would get a 401 with no redirect. I moved it into apiFetch() instead so all fetch-based calls get the redirect.

lynxkite-app/web/src/common.ts
```diff
@@ -111,6 +131,7 @@ export async function apiFetch(input: RequestInfo | URL, init?: RequestInit): Pr
   });
   if (response.status === 401) {
     await triggerLogin();
+    throw new Error("Unauthorized");
```
darabos, line 143: Is this needed? What does it change?

lynxkite-app/web/src/main.tsx
```diff
@@ -14,8 +14,15 @@ import {
...
   useRouteError,
 } from "react-router";
+import { useConfig } from "./common.ts";
 import Directory from "./Directory.tsx";
 
+function ConfigGuard({ children }: { children: ReactElement }) {
+  const config = useConfig();
+  if (!config.data) return null;
+  return children;
+}
```
darabos: Is this needed? Everything worked fine without it... It serializes all requests after the config request, which I don't think is necessary. Most of the components don't depend on config.

lynxkite-app/web/src/main.tsx
```diff
@@ -49,18 +56,45 @@ function WorkspaceError() {
...
+          <ConfigGuard>
+            <Directory />
+          </ConfigGuard>
+        }
+      />
       <Route path="/auth/callback" element={withSuspense(<AuthCallback />)} />
       <Route
         path="/edit/*"
-        element={withSuspense(<Workspace />)}
+        element={<ConfigGuard>{withSuspense(<Workspace />)}</ConfigGuard>}
```
darabos: Since these components already use Suspense, if we wanted to add a "ConfigGuard", what should probably also use Suspense. (But I hope we don't have to add it.)

lynxkite-app/web/src/common.ts
banhidiboti: It is needed. There are two separate HTTP stacks in use. Axios has its own interceptor for 401s, but that doesn't cover native fetch calls — which is what apiFetch uses.

lynxkite-app/web/src/main.tsx
banhidiboti: It is needed. I tested without it and the file list showed "Unauthorized" for several seconds. Without the guard, the dir list request fires before config loads, so triggerLogin silently does nothing on the first 401 and SWR's retry backoff is slow.

lynxkite-app/web/src/main.tsx
banhidiboti: Tried it but it made things "flashier". With Suspense, the config must resolve before lazy imports can start, so there are two sequential loading phases instead of one.

lynxkite-app/web/src/common.ts
darabos: > It is needed. There are two separate HTTP stacks in use. Axios has its own interceptor for 401s, but that doesn't cover native fetch calls — which is what apiFetch uses.

And how does this exception play into that?

lynxkite-app/web/src/main.tsx
darabos: > It is needed. I tested without it and the file list showed "Unauthorized" for several seconds. Without the guard, the dir list request fires before config loads, so triggerLogin silently does nothing on the first 401 and SWR's retry backoff is slow.

We could still wait for the config later. We don't need to hold up the page load when there is no 401. And even if there is authentication, we would save time by firing the config request and the directory listing request (which will 401) in parallel.

But okay, let's not make this more complicated than necessary. It makes everything simpler if we load the config at the start. Check out #254 — I think it achieves the same thing in a simpler way. No ConfigGuard, no useConfig.

lynxkite-app/web/src/common.ts
banhidiboti: Without the throw, apiJson would still try to process the 401 response while the redirect is happening. The throw stops that.

lynxkite-app/web/src/common.ts
```diff
@@ -143,6 +164,31 @@ export async function completeLoginCallback(): Promise<string> {
   return state?.returnTo || "/";
 }
 
+export function useConfig() {
+  return { data: cachedConfig };
+}
```
darabos: Delete this. Use `getConfig()` instead.

lynxkite-app/web/src/main.tsx
darabos: > But okay, let's not make this more complicated than necessary. It makes everything simpler if we load the config at the start. Check out #254 — I think it achieves the same thing in a simpler way. No ConfigGuard, no useConfig.

I see you've done this — updated this PR to use the solution from #254. But you didn't say anything about it. Please use GitHub comments for communication. Just pushing commits is not enough.

lynxkite-app/web/src/common.ts
banhidiboti: Done

banhidiboti: I can't comment on the comment above but I am sorry about the forgotten communaction before.

darabos: All good! Merged! 🚀  Thanks!

PR #241, opened 2026-06-04
First half of Soccer

KovacsHunor:
The first half of the Soccer(#240) demo workspace. I am not sure whether it was a good idea to add the csv files as well.


lynxkite-graph-analytics/src/lynxkite_graph_analytics/operations/basic_ops.py
```diff
@@ -57,6 +57,102 @@ def derive_property(
...
+    b.dfs[new_name] = b.dfs.pop(old_name)
+    return b
+
+
+@op("Use table as vertices", color="orange", icon="link")
+def use_as_vertices(b: core.Bundle) -> core.Bundle:
+    """Creates a graph with vertices from the table and with no edges"""
+    b = b.copy()
+    b.dfs["nodes"] = b.dfs.pop(list(b.dfs.keys())[0])
+    b.dfs["edges"] = pd.DataFrame()
```
darabos: This brings attention to an unsettled design issue.

In the old LynxKite a graph has nodes and edges and related graphs called segmentations. But my idea was that in LynxKite 2000 we would adopt a looser structure. The Bundle. A collection of tables, some of which describe nodes, edges, or other stuff. The semantics of the tables would be defined by the RelationDefinitions, which can tell us that the columns "user" and "product" in table "orders" describe edges between nodes described in the "users" and "products" tables.

That's the dream. It's a simple transition between loose tables imported from CSVs into a graph by just adding some metadata. In practice, when converting a NetworkX table into a Bundle we convert it into "nodes" and "edges" tables. So instead of looking at RelationDefinitions, it was often easier to just assume "nodes" is nodes and "edges" is edges. "We will fix it later."

This "Use table as vertices" box would make the temporary hack a permanent feature. Before we resort to this, let's try to give the dream a chance. Could you attempt to implement Soccer with RelationDefinitions instead of relying on table names? The Organize box is supposed to let you manipulate those RelationDefinitions, but I'm not sure it's powerful enough to do what you need.

lynxkite-graph-analytics/src/lynxkite_graph_analytics/operations/basic_ops.py
```diff
@@ -57,6 +57,102 @@ def derive_property(
...
+    asc = "ascending"
+    desc = "descending"
+
+
+@op("Add rank attribute", color="orange", icon="link")
+def add_rank(
+    b: core.Bundle,
+    *,
+    table_name: core.TableName,
+    attribute: core.ColumnNameByTableName,
```
darabos: For this combo we have a TableColumn type.

lynxkite-graph-analytics/src/lynxkite_graph_analytics/operations/basic_ops.py
```diff
@@ -57,6 +57,102 @@ def derive_property(
...
+        df,
+        df,
+        left_on=attribute1,
+        right_on=attribute2,
+        suffixes=("_src", "_dst"),
+    )
+
+    df_out = df_out[df_out["_id_src"] < df_out["_id_dst"]]
+    df_out = df_out.drop(columns=["_id_src", "_id_dst"])
+    b.dfs["edges"] = df_out
```
darabos: Plus a RelationDefinition!

lynxkite-graph-analytics/src/lynxkite_graph_analytics/operations/basic_ops.py
```diff
@@ -57,6 +57,102 @@ def derive_property(
...
+    return b
+
+
+@op("Connect vertices on attribute")
+def connect_vertices(
+    b: core.Bundle,
+    *,
+    table_name: core.TableName,
+    attribute1: core.ColumnNameByTableName,
+    attribute2: core.ColumnNameByTableName,
```
darabos: This could be meaningful between two different tables too. This is the main attraction of the RelationDefinition approach. We can have a "users" table and a "products" table that represent two different types of nodes, and we can have edges between them. A heterogeneous graph.

You could use two `TableColumn` parameters.

lynxkite-graph-analytics/src/lynxkite_graph_analytics/operations/basic_ops.py
```diff
@@ -57,6 +57,102 @@ def derive_property(
...
+) -> core.Bundle:
+    b = b.copy()
+    df = b.dfs[table_name].copy()
+
+    df["_id"] = range(len(df))
```
darabos: I think we could use `df.index` for this. Usually it's the same, the row index. But in a table of users the index could be the email address too, or whatever. If we add boxes that manipulate the index later, this box can be ready to roll with it!

lynxkite-graph-analytics/src/lynxkite_graph_analytics/operations/basic_ops.py
```diff
@@ -57,6 +57,102 @@ def derive_property(
...
+
+    df = df.sort_values(by=attribute, ascending=(order == OrderType.asc))
+    df[rank_name] = range(len(df))
+
+    b.dfs[table_name] = df
+    return b
+
+
+@op("Connect vertices on attribute")
+def connect_vertices(
```
darabos: In the old LynxKite we called them "vertices" but now we generally call them "nodes".
```suggestion
@op("Connect nodes on attribute")
def connect_nodes(
```

lynxkite-graph-analytics/src/lynxkite_graph_analytics/operations/basic_ops.py
```diff
@@ -57,6 +57,102 @@ def derive_property(
     return b
 
 
+@op("Filter with formula")
+def filter_with_formula(
+    b: core.Bundle, *, table_name: core.TableName, formula: ops.LongStr
+) -> core.Bundle:
+    """Removes all rows where the formula evaluates to false"""
```
darabos: We should link to some relevant documentation for the formula syntax. E.g. https://numexpr.readthedocs.io/en/latest/user_guide.html#supported-functions

darabos (commented):
Super nice work! Thank you!

Yeah, those CSVs are too big and I'm not even sure about their licensing status. Better not store them in the repo. The import boxes should be able to import from web URLs. I've uploaded these files to GCS now:

https://storage.googleapis.com/lynxkite_public_data/pzubcsek/Soccer/matches.csv
https://storage.googleapis.com/lynxkite_public_data/pzubcsek/Soccer/players.csv
https://storage.googleapis.com/lynxkite_public_data/pzubcsek/Soccer/events.csv

To avoid having these files in the Git history you have two options: use rebase or amend to edit them out of the history and then force push this branch. Or delete them with a new commit and merge this PR with a squash commit when done. (Let me know if I can help.)

lynxkite-graph-analytics/src/lynxkite_graph_analytics/operations/basic_ops.py
KovacsHunor: Done, I got rid of this function and used RelationDefinitions where it was relevant.

lynxkite-graph-analytics/src/lynxkite_graph_analytics/operations/basic_ops.py
KovacsHunor: Done

lynxkite-graph-analytics/src/lynxkite_graph_analytics/operations/basic_ops.py
KovacsHunor: Done

lynxkite-graph-analytics/src/lynxkite_graph_analytics/operations/basic_ops.py
KovacsHunor: I rewrote it to work with two different tables as well, but since for the RelationDefinitions we must know which columns can identify the nodes uniquely, I used 2 TableNames and 2-2 string for the IDs and attributes

lynxkite-graph-analytics/src/lynxkite_graph_analytics/operations/basic_ops.py
KovacsHunor: Done

lynxkite-graph-analytics/src/lynxkite_graph_analytics/operations/basic_ops.py
KovacsHunor: Done

lynxkite-graph-analytics/src/lynxkite_graph_analytics/operations/basic_ops.py
```diff
@@ -110,46 +102,75 @@ class OrderType(enum.StrEnum):
...
     *,
-    table_name: core.TableName,
-    attribute1: core.ColumnNameByTableName,
-    attribute2: core.ColumnNameByTableName,
+    table1: core.TableName,
+    id1: str,
+    attribute1: str,
+    table2: core.TableName,
+    id2: str,
+    attribute2: str,
```
darabos: The parameter names are visible on the UI, so try to make them human-friendly. Underscores get converted to spaces. How about `source_table`, `source_id`, `source_attribute`, `target_table`, etc?

And you can define a one-off parameter type for these right here, before the function.

```python
ColumnNameForSource = typing.Annotated[
    str, {"format": "dropdown", "metadata_query": "[].dataframes[].<source_table>.columns[]"}
]
ColumnNameForTarget = typing.Annotated[
    str, {"format": "dropdown", "metadata_query": "[].dataframes[].<target_table>.columns[]"}
]
```

darabos (approved):
Nice!

lynxkite-graph-analytics/src/lynxkite_graph_analytics/operations/basic_ops.py
KovacsHunor: Done

lynxkite-graph-analytics/src/lynxkite_graph_analytics/operations/basic_ops.py
```diff
@@ -57,6 +60,130 @@ def derive_property(
...
+    - source_id: ID column in the first table
+    - source_attribute: Attribute column in the first table used for matching
+    - target_table: Name of the second table
+    - target_id: ID column in the second table
+    - target_attribute: Attribute column in the second table used for matching
+    """
+    b = b.copy()
+
+    df1 = (b.dfs[source_table]).add_suffix("_src")
+    df2 = (b.dfs[target_table]).add_suffix("_dst")
```
darabos, line 156:
```suggestion
    df1 = b.dfs[source_table].add_suffix("_src")
    df2 = b.dfs[target_table].add_suffix("_dst")
```

darabos (approved):
> Well, this is quite strange. With my last commit the tests pass, even though I did not modify any of the boxes that `Model use.lynxkite.json` uses, nor `Model use.lynxkite.json` itself...

Probably just a timeout then! This is good to merge! 👍

PR #242, opened 2026-06-04
Adopt LIMs in open source

doragera:


lynxkite-graph-analytics/src/lynxkite_graph_analytics/elementwise.py
```diff
@@ -8,6 +8,22 @@
...
+    from lynxkite_enterprise.lim import (  # ty: ignore[unresolved-import]
+        call_lim_worker,
+        ensure_lim_worker_ready,
+        is_lim_worker,
+        lim_to_k8s_needs_kwargs,
+        make_service_name,
+    )
+    from lynxkite_enterprise import k8s  # ty: ignore[unresolved-import]
+
+    enterpise_backend = True
```
darabos:
```suggestion
    enterprise_backend = True
```

darabos (commented):
LGTM for the change in `main.py`. For `elementwise.py` I will have to take a closer look. I'm surprised it had to be changed at all. I thought there would be a function that it calls for every element. Whether that function happens to do the work locally, or sends it off to a microservice, why does that matter for `elementwise`?

lynxkite-app/src/lynxkite_app/crdt_update.py
```diff
@@ -138,25 +138,25 @@ def _crdt_update_list_by_index(
             # Update in place.
             if isinstance(value, dict):
                 if not isinstance(crdt_obj[i], pycrdt.Map):
-                    crdt_obj[i] = pycrdt.Map()  # ty: ignore[invalid-assignment]
+                    crdt_obj[i] = pycrdt.Map()
```
darabos, line 141: Nice!

lynxkite-graph-analytics/src/lynxkite_graph_analytics/elementwise.py
```diff
@@ -26,8 +39,19 @@ def elementwise(
 
     Example:
         @elementwise(input_table="protein_table_column", desc="Processing proteins")
-        async def process(record: Record, *, protein_table_column: tuple[str, str]):
+        async def process(record: Record, *, protein_table_column: TableColumn):
             record.result = await api_call(record[protein_table_column])
```
darabos: Please also add `@op` to this example, to make it clear which order the decorators should be applied.

lynxkite-graph-analytics/src/lynxkite_graph_analytics/elementwise.py
```diff
@@ -26,8 +39,19 @@ def elementwise(
...
+
+    With LIM, pass a small ``lim_config``; storage and k8s defaults are filled in
+    automatically::
```
darabos: Let's add one more sentence to make it clear that this is an Enterprise feature, and briefly what it does. E.g.:

```suggestion
    If LynxKite 2000:MM Enterprise is installed, you can run elementwise operations on Kubernetes
    as LIMs. With LIM, pass a small ``lim_config``; storage and k8s defaults are filled in
    automatically::
```

lynxkite-graph-analytics/src/lynxkite_graph_analytics/elementwise.py
```diff
@@ -26,8 +39,19 @@ def elementwise(
...
+
+        @elementwise(
+            input_table="protein_table_column",
+            desc="ESM2",
+            lim_config={"cpu": "4", "memory": "8Gi"},
+        )
+        async def query_esm2(record: Record, *, protein_table_column: TableColumn):
+            record.embeddings = my_inference(record[protein_table_column])
```
darabos: I think we should really support sync functions. My arguments for sync:
- LIMs are a solution when you realize that your code is slow. Slow code is typically sync.
- Inference code for random HF models is typically sync.
- Even your example is sync! You're not awaiting `my_inference()`.

lynxkite-graph-analytics/src/lynxkite_graph_analytics/elementwise.py
```diff
@@ -94,39 +125,83 @@ async def _elementwise_impl_async(
...
+            process_row_updates = remote_processor  # type: ignore[assignment]
+            lim_cleanup = getattr(remote_processor, "lim_cleanup", None)
+
+    try:
+        await _process_rows_concurrently(
+            self,
+            df,
+            process_row_updates,
+            desc=desc,
+            parallelism=execution_parallelism(self) if execution_parallelism else 1,
```
darabos: We could still keep the `concurrency` parameter. If I have a simple function that calls a web service there's no reason to run that in a LIM, but it would be pretty nice to run it with concurrency!

lynxkite-graph-analytics/src/lynxkite_graph_analytics/elementwise.py
```diff
@@ -94,39 +125,83 @@ async def _elementwise_impl_async(
...
+        )
+    finally:
+        if lim_cleanup is not None:
+            await lim_cleanup()
```
darabos, line 193: If it wasn't for this, I would say the whole LIM code could be a decorator in `lynxkite_enterprise` and `elementwise.py` wouldn't need to know about it. `get_remote_row_processor()` is almost that decorator.

But it's fine for now! Let's just get the functionality in, then we can figure out the best interface for setting up LIMs.

Just to float an idea: we could make it a configuration choice. `elementwise()` would consult a configuration, and if the box name is on the LIM list, it gets run as a LIM, otherwise it runs as a normal function.

darabos (approved):
Awesome, thank you!

darabos (approved):

Issue #237, opened 2026-05-28
CRDT resize updates are slow/delayed


darabos:
I resized a box up and down while Gabor B was looking at the same workspace on HF. My resizing was replayed for him in slow motion over minutes after I did it. 😐 

Possible explanations (without having looked at anything):
- The updates are big. The updates should be tiny for a resize event, but maybe something goes wrong and they get huge? The workspace had a bunch of large visualizations.
- There are too many updates. I think it's limited by frame rate, so 60 updates per second? Is that too many? We could try throttling it and only updating the CRDT size 5 times per second.
- The backend forwards the events slowly. Maybe the backend does too much work for these insignificant updates. Even just checking the workspace against the last version may be fairly expensive. **I think this is the likely explanation.** We could fix this by looking at the CRDT changes in `workspace_changed()` before even doing `ws_pyd = workspace.Workspace.model_validate(ws_crdt.to_py())`. This line is already expensive if the workspace weighs multiple megabytes. (#120) We can avoid parsing the workspace if we see that the change is just a resize.
- If all the above are false, we could have a continuous flow of tidy small updates arrive in Gabor's browser. And then his frontend could be overwhelmed and apply those updates very inefficiently.

@Mackofi This sounds like an issue you may enjoy? (Or not, if you've had enough of CRDT already.)

PR #243, opened 2026-06-05
Removing displays from CRDT and other optimizations.

### LINKED ISSUE #120: Move large displays outside of CRDT
It's convenient to make the display outputs (images, visualizations, tables) part of CRDT. But images (base64 encoded!) are pretty large and it's not efficient to store and share with the frontend their whole history. Let's figure out a better system.

Instead of `data:` images, for example, the display could be an `https:` image, fetched with a separate request to the backend. What we lose with the extra request we surely gain back by not fetching all the past images as well.

Mackofi:
Long awaited improvement that moves large displays (e.g. images, giant tables) from the CRDT backend to seperate workspace files (#120). The idea is the following:
- We save each node's display into their own seperate json file whenever we detect change.
- The frontend queries if the node has any display saved, if they do then fetches the json based on workspace and node ID. (We could maybe save the file location into the CRDT, but this might be cleaner?)
- We now took the display out of the CRDT, possibly saving 100s of MBs of website data that we would have to send to every client that joins leading to possible performance improvements.

Currently in draft because:
- I am not satisfied with the current solution of calling `updateState` every time.
- Performance feels a bit better, have not measured it.
- Possibly full of bugs 😄, and needs cleanup.
- Not yet sure on if this makes #237 irrelevant.

I just wanted to get the general idea across, so we could iterate on this. Anyone has any suggestions they are welcome!


lynxkite-app/src/lynxkite_app/crdt.py
```diff
@@ -168,7 +167,46 @@ def on_change(changes):
         return room
 
 
-last_ws_input = None
+def _workspace_fingerprint_from_dict(ws_dict):
+    """Produce a lightweight fingerprint for change detection.
+
+    Operates on a plain dict (e.g. from CRDT .to_py() or model_dump),
+    avoiding expensive model_validate, deep copy, and pydantic __eq__.
```
darabos, line 242: Nice!

lynxkite-app/src/lynxkite_app/crdt.py
```diff
@@ -264,17 +302,19 @@ async def workspace_changed(name: str, changes: list[pycrdt.MapEvent], ws_crdt:
         changes: Changes performed to the workspace.
         ws_crdt: CRDT object representing the workspace.
     """
-    ws_pyd = workspace.Workspace.model_validate(ws_crdt.to_py())
+    # if all(all(key in ("height", "width") for key in change.keys) for change in changes):
+    #     # Ignore changes that only update the height and width, which are triggered by the frontend when moving nodes around.
+    #     print("Ignoring change with keys", [change.keys for change in changes])
+    #     return
```
darabos: Can be probably deleted?

lynxkite-app/src/lynxkite_app/crdt.py
```diff
@@ -264,17 +302,19 @@ async def workspace_changed(name: str, changes: list[pycrdt.MapEvent], ws_crdt:
...
+    raw = ws_crdt.to_py()
+    ws_fingerprint = _workspace_fingerprint_from_dict(raw)
+    if ws_fingerprint == last_known_versions.get(name):
+        return
```
darabos: Generally we still want to save the workspace here. Maybe you only reorganized the boxes and wrote a bunch of comments. We don't want to redo the calculations. But we should still update the `lynxkite.json`.

If you add the logic for this, please add a comment about saving. It's easy to overlook this.

lynxkite-core/src/lynxkite_core/workspace.py
```diff
@@ -110,7 +112,33 @@ def publish_result(self, result: ops.Result):
             with nc.doc.transaction():
                 try:
                     nc["data"]["status"] = NodeStatus.done
-                    nc["data"]["display"] = self.data.display
+                    # If old version had display and new run has None,
+                    # frontend keeps previous cached display keyed by unchanged version.
+                    # TODO: check if this is an actual problem.
+                    #   (e.g. the boxes that use display always return something)
+                    if self._ws_path and result.display is not None:
+                        path = f".workspace_files/{self._ws_path.removesuffix('.lynxkite.json')}/{self.id}.json"
```
darabos: If the workspace lives in a subdirectory, could you also put the output files (`.workspace_files`) into that subdirectory? That way if we delete a directory full of old workspaces we also clean up their outputs.

Likewise `delete_workspace()` in `main.py` could delete the output of the workspace.

lynxkite-core/src/lynxkite_core/workspace.py
```diff
@@ -73,6 +74,7 @@ class WorkspaceNode(BaseConfig):
     width: Optional[float] = None
     height: Optional[float] = None
     _ws_crdt: Optional["pycrdt.Map"] = None
+    _ws_path: Optional[str] = None
```
darabos: The `Workspace` object knows the path already. But I see there's no way to get at that from `publish_result()`... Anyway, it would be nice to keep `publish_result()` simple.

How about putting the file creation logic in a new method on `ops.Result`? It could be triggered by `execute()` before `publish_result()`. `execute()` has access to the `Workspace` and so has the path.

darabos (commented):
Nice! 👍 

I see the advantage in putting all outputs into JSON files. I was thinking that visualizations could be just regular PNG files, etc. They would be a bit smaller and a bit more normal. (E.g. you could put the URL in an email.) But now that I see this, I think it's not worth the effort. JSON is fine.

lynxkite-app/src/lynxkite_app/crdt.py
Mackofi: Yes, when I finalize this PR, I will take out all unnecessary comments and prints left from debugging and trial-and-error.

lynxkite-app/src/lynxkite_app/crdt.py
Mackofi: I was thinking on how to solve this, as saving 10-20 times a second whenever a user was dragging or resizing boxes was expensive, and I decided on de-bouncing the save itself, before we check the `ws_fingerprint`. Now we save when two seconds have passed without any changes (can be modified). Hopefully in large workspaces the I/O access times will improve.

lynxkite-core/src/lynxkite_core/workspace.py
Mackofi: Done! 
We could also look at downloading or uploading workspaces with these output files. Maybe we could automatically ZIP them when the user downloads a workspace? However accepting ZIP files from user input can lead to ZIP bombs...

lynxkite-core/src/lynxkite_core/workspace.py
Mackofi: I extended the `ops.Result` class and I put the call into `ops.Op.__call__`. I decided against `execute()` as we would need to modify every executor (if I was looking at the right `execute()`. In `Op.__call__`, `OpContext` provides us all of the useful information that we need.

lynxkite-app/src/lynxkite_app/crdt.py
darabos: > I was thinking on how to solve this, as saving 10-20 times a second whenever a user was dragging or resizing boxes was expensive, and I decided on de-bouncing the save itself, before we check the `ws_fingerprint`. Now we save when two seconds have passed without any changes (can be modified). Hopefully in large workspaces the I/O access times will improve.

Oh nice! Thank you, de-bouncing is really the way to go. I would perhaps even move de-bouncing up to the frontend. Creating a million CRDT updates as a box is moved is bad in the first place, even if the backend can handle it. We could de-bounce the events from ReactFlow so we convert them to CRDT events in bigger steps. (Sorry for suggesting this after the backend de-bouncing is already done. We could have both? In a separate PR?)

lynxkite-app/src/lynxkite_app/crdt.py
```diff
@@ -21,6 +23,20 @@
 
 router = fastapi.APIRouter()
 main_loop = None
+WORKSPACE_CHANGED_THROTTLE_SECONDS = 1.0
+
+
+@dataclass
+class WorkspaceRuntimeState:
```
darabos, line 30: Neat! 👍 Maybe call it WorkspaceRuntimeState**s**, to indicate that it's not a per-workspace object. Or you could actually make it a per-workspace object, and then just have a `state: dict[str, WorkspaceRuntimeState]`.

lynxkite-app/src/lynxkite_app/crdt.py
```diff
@@ -256,6 +314,23 @@ def print_diff(old, new, prefix=""):
...
+    """Save the workspace with the given name and contents. This is debounced to avoid excessive saves during rapid changes."""
+    this_task = asyncio.current_task()
+    ws_pyd = None
+    await asyncio.sleep(2)  # We only try to save every 2 seconds.
+    try:
+        path = pathlib.Path() / name
+        assert path.is_relative_to(pathlib.Path()), f"Path '{path}' is invalid"
+        ws_pyd = workspace.Workspace.model_validate(raw)
+        ws_pyd.save(path)
+    except asyncio.CancelledError:
```
darabos: There's no `await` in the `try` block, so I don't think we can get cancelled. The `try` should probably go around `sleep`.

lynxkite-app/src/lynxkite_app/crdt.py
```diff
@@ -264,19 +339,25 @@ async def workspace_changed(name: str, changes: list[pycrdt.MapEvent], ws_crdt:
         changes: Changes performed to the workspace.
         ws_crdt: CRDT object representing the workspace.
     """
-    ws_pyd = workspace.Workspace.model_validate(ws_crdt.to_py())
+    raw_obj = ws_crdt.to_py()
+    if not isinstance(raw_obj, dict):
+        return
```
darabos: When does this happen? (Maybe add a comment.)

lynxkite-app/src/lynxkite_app/crdt.py
```diff
@@ -369,10 +450,18 @@ async def lifespan(app):
...
+        del state.delayed_workspace_changes[name]
+    state.pending_workspace_changes.pop(name, None)
+    state.next_allowed_workspace_flush_at.pop(name, None)
+    if name in state.delayed_executions:
+        state.delayed_executions[name].cancel()
+        del state.delayed_executions[name]
+    if name in state.delayed_saves:
+        state.delayed_saves[name].cancel()
+        del state.delayed_saves[name]
+    state.last_known_versions.pop(name, None)
```
darabos: If WorkspaceRuntimeState were a per-workspace object, it could have a `destroy()` method that handles all this.

lynxkite-core/src/lynxkite_core/ops.py
```diff
@@ -196,6 +198,16 @@ class Output(BaseConfig):
...
 
 
+def build_output_path(ws_path: str, node_id: str):
+    """Builds the path to save the display data for a given node output."""
+    if "/" in ws_path:
+        # If the ws is in a subdirectory, we save to that subdirectory
+        ws_path_obj = pathlib.Path(ws_path)
+        return f"{ws_path_obj.parent}/.workspace_files/{ws_path_obj.stem.removesuffix('.lynxkite')}/{node_id}.json"
+    else:
+        return f".workspace_files/{ws_path.removesuffix('.lynxkite.json')}/{node_id}.json"
```
darabos: It looks correct, but we could keep it simpler.

```suggestion
    ws_path_obj = pathlib.Path(ws_path)
    return f"{ws_path_obj.parent}/.workspace_files/{ws_path_obj.name}/{node_id}.json"
```

Yeah, this will result in `./` at the start for files in the root, but that should be fine. (If not, just add `.removeprefix("./")`.)

I also don't see a reason to remove `.lynxkite.json`. Nobody sees these paths. And maybe we will have other workspace types in the future. Having the full path can only be good.

lynxkite-core/src/lynxkite_core/ops.py
```diff
@@ -209,10 +221,34 @@ class Result:
...
+    def save_display(self, path: str | None, node_id: str):
+        """Saves the display data to a file. The path is relative to the workspace's data directory."""
+        # If old version had display and new run has None,
+        # frontend keeps previous cached display keyed by unchanged version.
+        # TODO: check if this is an actual problem.
+        #   (e.g. the boxes that use display always return something)
+        if path and self.display is not None:
+            path = build_output_path(path, node_id)
+            dirname, basename = os.path.split(path)
+            if dirname:
```
darabos: We don't even need this `if` if we don't remove `./`.

Maybe it's best to return a Path from `build_output_path()`. Then this could just be `path.parent.mkdir(exist_ok=True, parents=True)`.

lynxkite-core/src/lynxkite_core/ops.py
```diff
@@ -272,19 +309,29 @@ def __call__(self, op_ctx: OpContext, *inputs, **params):
...
-                "molecule",
-            ]:
+            if is_visualization_type:
                 # If the operation is a visualization, we use the returned value for display.
                 res = Result(display=res)
             else:
                 res = Result(output=res)
+
+        # Save display if needed
+        if is_visualization_type:
```
darabos: I don't think this check is needed here. If there's no display, `save_display` does nothing. I don't think this needs to depend on the box type.

darabos (commented):
Thank you!

lynxkite-app/src/lynxkite_app/crdt.py
Mackofi: I decided to not de-bounce saving, as if we only trigger changes every second, saving should not be an issue.

lynxkite-app/src/lynxkite_app/crdt.py
Mackofi: Co-pilot's overly secure coding habits made that change. This probably does not happen ever, as it was never in the code and we did not have errors.

lynxkite-core/src/lynxkite_core/ops.py
Mackofi: I removed the `.lynxkite.json` because of some file system troubles, but that is seemingly fixed(?) and was unrelated. Change accepted!

lynxkite-core/src/lynxkite_core/ops.py
Mackofi: Node types like `NodeWithGradio` uses the display attribute for something a bit different from what I could gather. Should we also make those call to the backend no matter what we write into `display`?

darabos: > Node types like `NodeWithGradio` uses the display attribute for something a bit different from what I could gather. Should we also make those call to the backend no matter what we write into `display`?

I see what you mean. Yeah, go with whatever is easiest. The Gradio `display` only contains a URL, so any solution will be fine.

Mackofi: Well we would not need to check the type of node, and save the display always, but for every node that uses `display` we would need a backend request to a file. But that may be a good convention to always call the backend for a node's `display` attribute. Currently it's all over the place... ~~I will change it to always save the display.~~

lynxkite-app/src/lynxkite_app/crdt.py
```diff
@@ -21,6 +23,28 @@
...
+        self.last_known_version = None
+        if self.delayed_workspace_change:
+            self.delayed_workspace_change.cancel()
+        self.pending_workspace_changes = []
+        self.next_allowed_flush_at = 0.0
+        if self.delayed_execution:
+            self.delayed_execution.cancel()
+
+
+state: dict[str, WorkspaceRuntimeState] = {}
```
darabos, line 47: 👍

lynxkite-app/web/src/workspace/crdt.ts
```diff
@@ -253,6 +264,13 @@ class CRDTConnection {
         console.log("Unknown node change", ch);
       }
     }
+
+    // Remove items in reverse order to avoid index shifting issues
+    for (let i = removeOperations.length - 1; i >= 0; i--) {
+      const { index } = removeOperations[i];
+      wnodes.delete(index);
+    }
```
darabos, line 273: The order in changes is not necessarily in order of increasing index. I think you need to sort `removeOperations` before this loop.

lynxkite-app/web/src/workspace/crdt.ts
```diff
@@ -264,16 +282,22 @@ class CRDTConnection {
...
         const edgeIndex = wedges.map((n: Y.Map<any>) => n.get("id")).indexOf(ch.id);
-        wedges.delete(edgeIndex);
+        removeOperations.push({ index: edgeIndex });
         wsChanged = true;
       } else if (ch.type === "select") {
       } else {
         console.log("Unknown edge change", ch);
       }
     }
+    // Remove edges in reverse order to avoid index shifting
```
darabos, line 297: Same.

darabos (approved):
Epic! One minor comment but otherwise good to merge! Thank you!

lynxkite-app/web/src/workspace/crdt.ts
Mackofi: Good catch!

PR #244, opened 2026-06-08
File upload

banhidiboti:
Users can now upload files directly from the directory browser. Files can be uploaded via a new "Upload file" button in the action bar, or by dragging and dropping them onto the file listing.

### DIFF
--- a/lynxkite-app/src/lynxkite_app/main.py
+++ b/lynxkite-app/src/lynxkite_app/main.py
@@ -5,6 +5,7 @@ import pydantic
 import fastapi
 import joblib
 import pathlib
+import starlette.datastructures
 from fastapi.staticfiles import StaticFiles
 from fastapi.middleware.gzip import GZipMiddleware
 import starlette.exceptions
@@ -171,14 +172,16 @@ async def service_post(req: fastapi.Request, module_path: str):
 
 
 @app.post("/api/upload")
-async def upload(req: fastapi.Request):
-    """Receives file uploads and stores them in DATA_PATH."""
-    await auth.check_permission(req, "write", "uploads")
+async def upload(req: fastapi.Request, dir: str = "uploads"):
+    """Receives file uploads and stores them in DATA_PATH/dir."""
+    await auth.check_permission(req, "write", dir)
+    upload_dir = data_path / dir
+    assert upload_dir.is_relative_to(data_path), f"Path '{upload_dir}' is invalid"
     form = await req.form()
     for file in form.values():
-        if not isinstance(file, fastapi.UploadFile) or not file.filename:
+        if not isinstance(file, starlette.datastructures.UploadFile) or not file.filename:
             continue
-        file_path = data_path / "uploads" / file.filename
+        file_path = upload_dir / file.filename
         assert file_path.is_relative_to(data_path), f"Path '{file_path}' is invalid"
         with file_path.open("wb") as buffer:
             shutil.copyfileobj(file.file, buffer)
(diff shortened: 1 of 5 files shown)


lynxkite-app/src/lynxkite_app/main.py
```diff
@@ -171,14 +172,16 @@ async def service_post(req: fastapi.Request, module_path: str):
 
 
 @app.post("/api/upload")
-async def upload(req: fastapi.Request):
-    """Receives file uploads and stores them in DATA_PATH."""
-    await auth.check_permission(req, "write", "uploads")
+async def upload(req: fastapi.Request, dir: str = "uploads"):
```
darabos, line 175: In general it's weird to have some parameters in the request body and others in the URL. For POST requests we generally put everything in the body. But in this case the request body is the multipart form data with the files. It's easier to leave that alone and put the directory name in the URL.

👍

lynxkite-app/src/lynxkite_app/main.py
```diff
@@ -171,14 +172,16 @@ async def service_post(req: fastapi.Request, module_path: str):
...
+    """Receives file uploads and stores them in DATA_PATH/dir."""
+    await auth.check_permission(req, "write", dir)
+    upload_dir = data_path / dir
+    assert upload_dir.is_relative_to(data_path), f"Path '{upload_dir}' is invalid"
     form = await req.form()
     for file in form.values():
-        if not isinstance(file, fastapi.UploadFile) or not file.filename:
+        if not isinstance(file, starlette.datastructures.UploadFile) or not file.filename:
```
darabos, line 182: Why did this have to change?

darabos (commented):
Thank you! The test is failing for some reason.

A new test would be nice for this! It's the kind of rarely-used feature that we can break accidentally and then not notice for months. 😅 Testing the drag&drop may be hard, but I think Playwright can do the click&pick option.

lynxkite-app/web/tests/upload.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+    buffer: Buffer.from("hello from upload test"),
+  });
+
+  await expect(splash.getEntry(fileName)).toBeVisible();
+
+  await splash.deleteEntry(fileName);
+  await expect(splash.getEntry(fileName)).not.toBeVisible();
+});
+
+test("Can upload multiple files at once", async ({ page }) => {
```
darabos, line 24: Wow, I didn't realize we supported this! Way cool!

lynxkite-app/web/tests/lynxkite.ts
```diff
@@ -182,14 +182,17 @@ export class Workspace {
       this.page.locator(`.react-flow__edge[aria-label="Edge from ${sourceId} to ${targetId}"]`),
     ).toBeAttached({ timeout: 1000 });
   }
-  async connectBoxes(sourceId: string, targetId: string) {
-    // The method above is unreliable. I gave up after a lot of debugging and added these retries.
-    while (true) {
+  async connectBoxes(sourceId: string, targetId: string, maxRetries = 20) {
+    // Drag-and-drop in ReactFlow is flaky, so we retry. Cap retries to avoid infinite hangs.
+    for (let i = 0; i < maxRetries; i++) {
```
darabos: What's up with this? Your test doesn't even open a workspace. Why change this?

Also I don't think the retries should be capped. The test will fail with a timeout at some point anyway. What do we gain by failing sooner?

darabos (commented):
Thank you for the tests!!!

lynxkite-app/web/tests/lynxkite.ts
banhidiboti: You right, the change was unrelated to the upload feature and slipped in by mistake. Reverted to the original

darabos (approved):
All good! Thank you!

lynxkite-app/web/tests/lynxkite.ts
banhidiboti: Turns out we need it. The cap matters for cleanup without it, timed-out tests leave the backend in a bad state and break the upload tests that run after.

darabos (approved):
Thank you! 🚀

PR #247, opened 2026-06-10
Tagline changed to Discover Drugs, Graphs and the World!

banhidiboti:


darabos (approved):
As a non-native speaker I am not qualified to assess whether this would be understood in the pharmacological or the recreational sense. Let's do it! 🚀

PR #249, opened 2026-06-11
Operation categorization

### LINKED ISSUE #82: Put existing operations into categories
We have operation categories now, but we didn't move existing operations into them. This change breaks compatibility, so let's do it sooner rather than later!

KovacsHunor:
Most operations are in categories now.
Resolves https://github.com/lynxkite/lynxkite-2000/issues/82


darabos (approved):
Thank you!!!

PR #251, opened 2026-06-12
Relations

KovacsHunor:
Making the bundle-graph conversion and graph visualization use RelationDefinitions

### DIFF
--- a/lynxkite-graph-analytics/src/lynxkite_graph_analytics/operations/graph_ops.py
+++ b/lynxkite-graph-analytics/src/lynxkite_graph_analytics/operations/graph_ops.py
@@ -28,8 +28,10 @@ def merge(
 @op("Define Edges", view="graph_creation_view", outputs=["output"], icon="link")
 def define_edges(b: core.Bundle, *, relations: str = ""):
     """Define edges between node tables"""
+    b = b.copy()
     if relations.strip():
-        b.relations = [core.RelationDefinition(**r) for r in json.loads(relations).values()]
+        new_relations = [core.RelationDefinition(**r) for r in json.loads(relations).values()]
+        b.relations.extend(new_relations)
     return ops.Result(output=b, display=b.to_table_view(limit=100))
 
 
--- a/lynxkite-graph-analytics/src/lynxkite_graph_analytics/operations/visualization_ops.py
+++ b/lynxkite-graph-analytics/src/lynxkite_graph_analytics/operations/visualization_ops.py
@@ -46,27 +46,27 @@ def visualize_graph(
     label_by: core.NodePropertyName | None = None,
     color_edges_by: core.EdgePropertyName | None = None,
 ):
-    if graph.relations:
-        node_tables = []
-        seen_tables = set()
+    node_tables = []
+    seen_tables = set()
 
-        def add_table(table: str, id: str):
-            if table not in seen_tables:
-                df = graph.dfs[table].copy()
-                df["_id"] = df[id]
-                df["_table"] = table
-                node_tables.append(df)
-                seen_tables.add(table)
+    def add_table(table: str, id: str):
+        if table not in seen_tables:
+            df = graph.dfs[table].copy()
+            df["_id"] = df[id]
+            df["_table"] = table
+            node_tables.append(df)
+            seen_tables.add(table)
 
-        for relation in graph.relations:
-            add_table(relation.source_table, relation.source_key)
-            add_table(relation.target_table, relation.target_key)
-        nodes_df = pd.concat(node_tables, ignore_index=True)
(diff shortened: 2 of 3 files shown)


lynxkite-graph-analytics/src/lynxkite_graph_analytics/operations/graph_ops.py
```diff
@@ -28,8 +28,10 @@ def merge(
 @op("Define Edges", view="graph_creation_view", outputs=["output"], icon="link")
 def define_edges(b: core.Bundle, *, relations: str = ""):
     """Define edges between node tables"""
+    b = b.copy()
```
darabos, line 31: Good catch!

Gergő and I already spent some time thinking about how we could defend against such mistakes.

I looked a bit more into how we could do this with a type check. I.e. operation functions could take a FrozenBundle as the input instead of a Bundle, and FrozenBundle wouldn't have any in-place modification methods. Also `dfs` would be `dict[str, FrozenDataFrame]`. FrozenBundle could be a Protocol that is a subset of the Bundle interface. So it would still receive a normal Bundle, but during type checking we would not have access to the mutating methods.

This seems easier said than done. https://github.com/terrier-org/pyterrier/issues/515 has some links for attempts at FrozenDataFrame.

In the end these options seem realistic to me:
- Give up on FrozenDataFrame but implement FrozenBundle. Hope that this will serve as a constant reminder to make copies.
- Copy bundles in the executor. Instead of trying to police box authors, this would solve the problem once and for all in a central location. Box code would become simpler.

This second option is really attractive, except that DataFrames can be big and if we need to copy them for every box it could take a lot of time and memory. But Pandas 3.0 uses [copy-on-write](https://pandas.pydata.org/docs/user_guide/copy_on_write.html). We can probably implement a fast shallow copy for Bundles that turns into a more expensive copy only when needed.

Implementing this is outside of the scope of this PR, but if you're interested, it could be in the scope of a separate PR!

lynxkite-graph-analytics/src/lynxkite_graph_analytics/operations/visualization_ops.py
```diff
@@ -71,11 +92,33 @@ def visualize_graph(
             curveness = 0  # Street maps are better with straight streets.
             break
     else:
+        print(nodes.columns)
```
darabos: Remove.

darabos (commented):
Very nice!

Would it be possible to move the node/edge collecting code into a function? That way some code could be shared between the normal and the binned graph. But more importantly, I expect these functions will become more complex over time. Now you simply get all relations. But later on we will want to user to have control over which relations are displayed.

Thank you!

lynxkite-graph-analytics/src/lynxkite_graph_analytics/operations/graph_ops.py
KovacsHunor: Okay, sounds like a good idea!

lynxkite-graph-analytics/src/lynxkite_graph_analytics/operations/visualization_ops.py
KovacsHunor: Good catch! Removed.

lynxkite-graph-analytics/src/lynxkite_graph_analytics/operations/visualization_ops.py
```diff
@@ -38,14 +38,7 @@ def _map_color(value):
...
 
-@op("Visualize graph", view="visualization", icon="eye", color="blue")
-def visualize_graph(
-    graph: core.Bundle,
-    *,
-    color_nodes_by: core.NodePropertyName | None = None,
-    label_by: core.NodePropertyName | None = None,
-    color_edges_by: core.EdgePropertyName | None = None,
-):
+def _nodes_and_edges(graph):
```
darabos: Please add a short docstring.

darabos (approved):
> Yes, I moved the collection into its separate function

Perfect, thank you! 👍 

> Although I just wanted the visualizations on the main branch to use relations, depending on what plans Anna has, any more changes might be pointless.

Anna switched from the visualization task to the LynxKite assistant's development! So now you are free to implement anything on the visualization front.

I will be in the office tomorrow. If you will be as well, we could discuss the visualization plans.

> By the way, is it intentional that in `visualize_graph` the y axis point to the opposite direction compared to `binned_graph_visualization`?

Haha, no! But I think we had them like this for 10+ years now. 😅 You're welcome to flip it around! Thanks!

lynxkite-graph-analytics/src/lynxkite_graph_analytics/operations/visualization_ops.py
KovacsHunor: Done

PR #254, opened 2026-06-17
Load the config at startup so it is always available

darabos:
@banhidiboti has convinced me in #238 that it's easier for everyone if the config is loaded at the start before everything else. But if we do that, we can make a number of things simpler.


lynxkite-app/web/src/ProgressPageDemo.tsx
```diff
@@ -359,16 +359,14 @@ function Workspaces(props: { workspaces: any[]; setResources: (ws: any, resource
                       `(${ws.current_box_progress.done + currentBoxDemoCounter * (ws.resources.gpus ?? 0)}/${ws.current_box_progress.input_size})`}
                   </span>
                   {ws.current_box_progress.input_size && (
-                    <>
```
darabos, line 362: I removed the `<>` because I noticed Biome was emitting a warning for it. No relation to the rest of the PR.

banhidiboti (approved):

PR #256, opened 2026-06-18
Assistant skills and comments

### LINKED ISSUE #186: Update lynxkite-vibe to let the coding agent use existing LynxKite boxes
This should be a generic solution, probably a Claude skill. That way any coding agent can use LynxKite boxes. While we are creating a skill, it should also explain how to construct `main()`.

### LINKED ISSUE #203: Handle comment boxes in the assistant's view
The assistant sees comments in `workspace.py` like this:

```
lynxkite_core.ops.<lambda>(text='# Loading the data\nEach record is a miRNA–mRNA interaction.')
```

Instead we should render it as a Python comment, or just skip it. In either case it will be hard to restore them on the Python-to-workspace side.

annapanni:
Implements the following features for the LynxKite Assistant:
- Comment boxes in the workspace are now converted to python comments in the agent's representation of the workspace (#203), and new python comments are converted to comment boxes
- Created a skills for operations (#186)
    + operations defined in the same file are grouped together to form a single skill
    + added pre-commit to regenerate the skill descriptions if one of the files in it's module were changed
- Display OpenAI AuthenticationError properly on the UI (instead of Internal Server Error)


lynxkite-assistant/src/lynxkite_assistant/generate_skills_from_catalog.py
```diff
@@ -0,0 +1,137 @@
...
+
+if __name__ == "__main__":
+    if len(sys.argv) == 2:
+        output_path = sys.argv[1]
+        create_skills_from_catalog(output_path)
+    elif len(sys.argv) > 2:
+        output_path = sys.argv[1]
+        create_skills_from_catalog(output_path, sys.argv[2:])
+    else:
+        create_skills_from_catalog()
```
doragera, line 166: Please add a small example in a comment here on the usage of the script to understand the args

lynxkite-assistant/src/lynxkite_assistant/python_workspace_conversion.py
```diff
@@ -31,6 +31,33 @@ def python_to_workspace(code: str) -> workspace.Workspace:
     saved_values = {}
     box_x = {}
     box_y = {}
+    comments = []
+    comment_lines = []
+    next_line = 0
+    # gather multiline comments
```
doragera, line 37: I thought you said there is an ast library that can work with these comments. Or was it only for one-line comments?

doragera (approved):
Great!
For restructuring the skills I would try out the structure described here https://docs.langchain.com/oss/python/deepagents/skills, and then compare the performance of the two.

There is a conflict due to modifying the `visualization_graph` function which was improved by Hunor in the meantime. If you want to work on it too later put it in a different PR and rebase it on the current version.

lynxkite-assistant/src/lynxkite_assistant/python_workspace_conversion.py
annapanni: Yes, the library is ast-comments, which includes comments as ast nodes, but creates separate nodes for multi-line comments.
I could still use it, but then I would need to load all mulltiline comments into a single comment line.  I figured the agent would have an easier time generating multi-line comments like this so it would be worth it even if it made the code less readabale.

lynxkite-assistant/src/lynxkite_assistant/generate_skills_from_catalog.py
```diff
@@ -0,0 +1,147 @@
...
+    if changed_files:
+        # assuming plugin name is the first part of the path, and - can be replaced with _ in the module name
+        changed_plugins = list({f.split("/")[0].replace("-", "_") for f in changed_files})
+        _detect_certain_plugins(changed_plugins)
+    else:
+        ops.detect_plugins()
+    if not os.path.exists(output_path):
+        os.makedirs(output_path)
+    box_skills = get_box_skills()
+    package_skills = itertools.groupby(box_skills, key=lambda s: s["package"])
```
doragera: `itertools.groupby()` only groups together the adjacent items next to each other. If `box_skills` is not sorted then different op registration order can give you different results, and also you can lose some ops if the ops from one package are scattered in the list and not consecutive. I think this could cause the pre-commit hook to fail in CI

PR #280, opened 2026-07-06
Fix custom boxes being unknown

### LINKED ISSUE #66: Custom boxes are "unknown" sometimes
I think the custom box definitions are loaded too late. So after a restart the edges are missing because the unknown boxes have no inputs/outputs. But re-running the workspace fixes it.

Mackofi:
The workspace is loaded before we load user-scripts, making the metadata of custom boxes non-existent, so we perceive them as `Unknown operation.` If we load the scripts before we update the metadata, that fixes the boxes being unknown.

(Fixes #66)


darabos (approved):
Thank you so much!!!

darabos (approved):

PR #287, opened 2026-07-06
Fix disappearing edges when ungrouping

darabos:
Fixes #284. Sometimes position is a normal JavaScript object and `pos.x`/`pos.y` works. Sometimes it's a Y.Map where you need `pos.get("x")`/`pos.get("y")`. (Or `pos.toJSON()` to get a normal JavaScript object.) In the latter case the existing code didn't find the x/y coordinates and ended up with NaN for the edges' coordinates, effectively making the edges disappear.

As for why `position` is sometimes a JavaScript object, sometimes a Y.Map, I think it depends on the history of the object. If it was loaded from a lynxkite.json file, it's a Y.Map, if it was created on the frontend, it's a plain object. I've tried to fix this with the second commit. But I think we can still keep the first commit. It's easier to test.


KovacsHunor (approved):
Super, thanks! I don't see any serious harm in leaving the first commit.

PR #289, opened 2026-07-08
Add hierarchical settings in folders

doragera:
`settings.yaml` files store configuration for a folder. Settings are inherited by child folders, with child values overriding those from parent folders. These files will be used to store LIM configurations and are also a suitable place for ACLs for #215 and #216.

### DIFF
--- a/lynxkite-core/pyproject.toml
+++ b/lynxkite-core/pyproject.toml
@@ -7,6 +7,7 @@ requires-python = ">=3.11"
 dependencies = [
     "pydantic>=2.11.7",
     "pydantic-core>=2.46.4",
+    "pyyaml>=6.0.2",
 ]
 license = "AGPL-3.0-or-later"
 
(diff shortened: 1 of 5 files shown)


lynxkite-core/pyproject.toml
```diff
@@ -7,6 +7,7 @@ requires-python = ">=3.11"
 dependencies = [
     "pydantic>=2.11.7",
     "pydantic-core>=2.46.4",
+    "pyyaml>=6.0.2",
```
darabos: 6.0.3 is the most recent version.

lynxkite-graph-analytics/src/lynxkite_graph_analytics/elementwise.py
```diff
@@ -49,15 +50,12 @@ def elementwise(
...
+    resources per folder in ``settings.yaml``:
+
         @op("ESM2", slow=True)
-        @elementwise(
-            input_table="protein_table_column",
-            desc="ESM2",
-            lim_config={"cpu": "4", "memory": "8Gi"},
-        )
+        @elementwise(input_table="protein_table_column", desc="ESM2")
+        @lim
```
darabos, line 58: Nice! I hope eventually we won't even need `@lim`.

darabos (approved):
The most well-tested piece of code in LynxKite! 😄

PR #290, opened 2026-07-10
Assistant: chat history, example tasks and more

annapanni:
- generate sample interactions form the doc with example assistant tasks. Theses are saved under `examples/generated_samples` and have to be evaluated manually, closes #261 
- now we execute workspace on write (as long as it's not paused) - this enables the assistant to see any new errors it might have created. Hopefully this means the user doesn't have to prompt "fix it" so often
- refactored the assistant file structure: broke down large functions and organized them into separate files by functionality
- assistant chat history is now displayed and saved in the json + it can be cleared by a button. The issue (#262) intended to keep multiple sessions per workspace, but I did not think this necessary.
- added a skill file for custom box creation based on existing documentation, including the descriptions of some of the custom types (eg. Bundle, TableColumn etc.)
- In the style of coding agents, the user can now choose to include a reference to the boxes they have selected, closes #264 
- assistant can now see the results of the executions under `.workspace_files`, closes #257 
- `boxes.py` now has custom examples for the 'LynxKite Graph Analytics environment' #253 
- added a more detailed README, closes #295


lynxkite-assistant/src/lynxkite_assistant/instructions.py
```diff
@@ -0,0 +1,113 @@
...
+
+SYSTEM_PROMPT = """
+## Overview
+You are an assistant for the LynxKite no-code AI workflow builder.
+You have access to the following files, non of which are visible to the user:
+- /workspace.py: The Python representation of the workflow.
+- /boxes.py: The definitions of custom boxes in the workflow.
+- /layout.json: The layout of the boxes in the visual representation.
+- /errors.txt: The errors that occurred in the boxes during execution.
+- /.workspace_files/: The results of the executed boxes, such as View tables and View images.
```
doragera: Here it is already the virtual paths that the agent can operate with, but in `assistant.py` you route it without the dot:
```
routes["/workspace_files/"] = backends.FilesystemBackend(
            root_dir=str(workspace_files_path), virtual_mode=True
        )
```

For the virtual paths you can skip the dot.

lynxkite-assistant/src/lynxkite_assistant/assistant.py
doragera: I know this is not included in the PR, but might be worth switching to `agent.astream` to not block the main thread

lynxkite-assistant/src/lynxkite_assistant/instructions.py
```diff
@@ -0,0 +1,113 @@
+"""Instructions for the LynxKite assistant. These are used to generate the system prompt and file comments for the assistant."""
+
+SYSTEM_PROMPT = """
+## Overview
+You are an assistant for the LynxKite no-code AI workflow builder.
+You have access to the following files, non of which are visible to the user:
```
doragera: typo: none of which

lynxkite-app/web/src/workspace/Assistant.tsx
```diff
@@ -87,7 +126,19 @@ export default function Assistant(props: { workspace: string }) {
           event.preventDefault();
           const prompt = input.trim();
           if (!prompt || status !== "ready") return;
-          sendMessage({ text: prompt });
+          if (includeSelectedNodes && selectedNodeIds.length > 0) {
+            setMessages([
+              ...persistedMessages,
```
doragera: Might be better to use `messages` here? Or is there a reason to use persistedMessages for the box references?

doragera (approved):
Great!

Make sure that you don't leave stale chat history in frontend/backend. If LynxKite stops unexpectedly, or the backend has an error, or the user just leaves the workspace for a bit the chat history should be kept in a correct state, and can be reloaded when restarting LynxKite.
Now backend saves at stream start, before the agent responds and assistant reply only appended at stream end. If backend errors mid-stream, chat history can end with user message, or UI can show partial messages. Are we handling this? You can easily reproduce an error in the backend if you temporily delete your API key that causes an Authentication error. For seeing what happens if the reply stops mid-stream you may have to be more creative!

lynxkite-app/web/src/workspace/Assistant.tsx
annapanni: Both of them worked okay for me, but it does make more sense to use `messages` - I switched it.

PR #291, opened 2026-07-10
Workspace progress

doragera:
Resolves https://github.com/lynxkite/lynxkite-2000/issues/162

I had to move https://github.com/lynxkite/lynxkite-2000-enterprise/blob/main/lynxkite-enterprise/src/lynxkite_enterprise/progress_crdt.py to this open source repo so the progress of each workspace and the updates are in one place. The main /progress page still only in enterprise.

The issue described 3 levels of the implementation of the progress bar, this PR implements level 1+2.

This is located in the horizontal bar on top of the workspace next to the gpu setter.
image

### DIFF
--- a/lynxkite-app/web/src/workspace/Workspace.tsx
+++ b/lynxkite-app/web/src/workspace/Workspace.tsx
@@ -59,6 +59,8 @@ import NodeWithMolecule from "./nodes/NodeWithMolecule.tsx";
 import NodeWithParams from "./nodes/NodeWithParams";
 import NodeWithTableView from "./nodes/NodeWithTableView.tsx";
 import NodeWithVisualization from "./nodes/NodeWithVisualization.tsx";
+import { useWorkspaceProgress } from "./useWorkspaceProgress.ts";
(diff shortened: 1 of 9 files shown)


lynxkite-app/web/src/workspace/Workspace.tsx
```diff
@@ -104,6 +106,7 @@ function LynxKiteFlow() {
     .pop()!
     .replace(/[.]lynxkite[.]json$/, "");
   const crdt = useCRDTWorkspace(path);
+  const workspaceProgress = useWorkspaceProgress(path, Boolean(crdt?.ws));
```
darabos: Could this be moved into the WorkspaceProgress component? If it's in the Workspace, React has to re-render the whole Workspace on every progress update. If it's in WorkspaceProgress, we only re-render that small component.

darabos (approved):
Wow this looks so good! 😍

lynxkite-app/web/src/workspace/Workspace.tsx
doragera: Done!

PR #292, opened 2026-07-13
Ability to view trajectories with Mol*

Mackofi:
Just a quick and simple addition to our molecule viewer. With this we can now display trajectories too!


darabos (approved):
Wow it's that simple? Fantastic!

PR #293, opened 2026-07-13
Resizing matplotlib figures based on node size

annapanni:
Added resizing on execute based on node size in the matplotlib_to_image wrapper.
As long as the function has self as the first parameter to access OpContext, the resizing should happen automatically.
closes #285


lynxkite-core/src/lynxkite_core/matplotlib_to_image.py
```diff
@@ -44,6 +46,15 @@ def matplotlib_to_image(func):
     @functools.wraps(func)
     def wrapper(*args, **kwargs):
         func(*args, **kwargs)
+        self, *rest = args
+        if isinstance(self, OpContext):
+            # If the first argument is an OpContext, we can use its node to determine the size of the figure.
+            if self.node:
+                w, h = self.node.width, self.node.height - 100  # Subtract 100 pixels for the header
+            else:
+                w, h = 500, 450
```
doragera: Are these magic numbers the current fixed numbers for a matplotlib image? Just add a comment about this too.

doragera (approved):
Nice! Thanks for updating the examples too.

